PR #228, opened 2016-06-28
Little oversight in the french tutorial

kaiomar:
Hello,

There is a little oversight in the french tutorial
Part 5, Section VI, Step 2 : "Toujours à partir d'EmuNAND" 
There should be a "9" after EmuNAND so it should be "Toujours à partir d'EmuNAND9"

I know it's nothing but better change it I think ^^

Anyway thanks a lot for this amazing work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,1 @@
-### See https://github.com/Plailect/Guide/wiki
+### Voir https://github.com/Ginkyway/Guide/wiki/Accueil


Plailect:
WTF?

Why is this a pull request? Why are you changing the README as a part of this pull request? Why is this pull request totally unrelated to the comment that you made? Why are you telling me instead of Ginkyway? What is going on here?

PR #354, opened 2016-08-18
Add icon-magnet.gif / add magnet links to the guide

DanTheMan827:
I think a good addition would be to add magnet links in addition to just the torrent file.

It's not a requirement but sometimes magnet links can be more convenient since they open the torrent client directly instead of having to download the torrent file and then opening that.

You can add the magnet icon / link to the end of the filename like so:

    [Magnet Link](magnet-url-here)

Here is a file with all the corresponding magnet links
[magnet links.txt](https://github.com/Plailect/Guide/files/425327/magnet.links.txt)


Plailect:
Github flavored markdown blocks magnet URIs in both their markdown link and href forms, hence the use of .torrent files.

PR #402, opened 2016-08-27
Added rss feed and rss feed generator (python)

### LINKED ISSUE #369: [Enhancement] Torrent rss
So I am perfectly happy seeding the torrents for you and I was thinking it would be great if you could create an rss feed for people to drop in their torrent clients so that we can automatically download the torrents and then seed them without us having to pay attention to it.

james-d-elliott:
This enhancement closes #369 

Can provide documentation if necessary (it's relatively simple to use).


Plailect:
Please open a new PR, and fix the following issues:

>error on line 13 at column 96: Namespace prefix media on content is not defined
error on line 14 at column 27: Namespace prefix media on hash is not defined

Issue #673, opened 2016-11-01
Consider removing FBI Injection entirely


cheatfreak47:
For Installing Arm9LoaderHax:

It's possible to use the Homebrew Launcher version of FBI to install the CIA version of it, negating the need for FBI injections into Health & Safety.

---
EDIT: 
I took the liberty of partially rewriting this section of the guide to suit this change. This shortens up the section a lot, of course if I missed anything it should be added in. But this is just the jist of the revision, basically.

[code for full section here](http://pastebin.com/raw/XkrMY3Eg)
The more important stuff starts at section III, at step 6, continuing from step 5 here.
> 4. **If, before following this guide, you already had an EmuNAND setup and would like to move the contents of your previous EmuNAND to your new SysNAND CFW, now is the time to [follow Move EmuNAND](move-emunand).**
> 
> ##### Section IV - Finalizing setup
> 
> 1. Launch the browser, then open the browser settings.
> 2. Scroll to the bottom and Initialize Savedata. (it also may be called Clear All Save Data)
> 3. Navigate to `http://yls8.mtheall.com/3dsbrowserhax_auto.php`
>     + You can also load the page by scanning the following QR code (press (L + R) on the home menu, then tap the QR icon on the bottom screen)     
> ![browserhax_auto](https://yls8.mtheall.com/3dsbrowserhax_auto_qrcode.png)
>     + If you get an error, [follow this troubleshooting guide](troubleshooting#ts_browser)
> 4. Your console should load the homebrew menu
> 5. Select "DSP Dump"
> 6. Press (Start) when prompted to exit back to the homebrew launcher
> 7. Select "FBI"
> 8. Select "SD"
> 9. Select "cias"
> 10. Navigate to `FBI.cia` and press (A) to install
> 11. Navigate to `hblauncher_loader.cia` and press (A) to install
> 12. Navigate to `lumaupdater.cia` and press (A) to install
> 13. Press (B) twice to return to the menu, and then press start to exit back to the homebrew launcher
> 14. Press (Start) and then (A) to reboot back into the home menu.
> 15. Unwrap each of the new applications.
> 16. Update your 3DS to the latest firmware by going to System Settings, then "Other Settings", then going all the way to the right and using "System Update"
>   + **Yes this is safe, stop asking about it.**
>   + If this gives you an error, set your DNS settings to "auto"
[...]

Issue #672, opened 2016-11-01
Consider pointing toward the latest release of SafeA9LHInstaller


cheatfreak47:
Seems that it's quite a bit more stable with the recent change of what ram to use that Aurora Wright implemented. 

~~Also, I'm pretty sure you don't need to link separate builds for Old3DS/New3DS.~~
forgot about 2xrsa, but the New 3DS really ought to be using the latest build.

Ev1lbl0w:
New3DS SafeA9LHInstaller uses an OTPless method to install arm9loaderhax, which isn't supported on the Old3DS. This is why there are separate builds for types of console.

ihaveamac:
actually it's because the newer SafeA9LHInstaller removed 2xrsa support temporarily due to memory corruption issues I think.

Plailect:
https://github.com/Plailect/Guide/commit/fa627ea72c6c4917842c36f74d1ae7d59311a4ef

CyHo:
@Plailect , this link become 404, for New 3DS Compatible Release

Plailect:
whoops, shouldn't have the "tag" in the link

Issue #671, opened 2016-11-01
DSiware Downgrade tells you to download Universal Inject Generator


cheatfreak47:
It's not used at all ¯\_(ツ)_/¯

PR #674, opened 2016-11-01
Remove FBI Injection Entirely, Misc Small Changes

cheatfreak47:
This PR does 3 things
- Addresses #671, #672, and #673 all of which were issues I started.

The biggest change here is FBI Injection removed.

I more or less restructured the last two sections of the previous page into one slightly larger section, but removed all complications involving FBI Injection, including the torrent, and magnets. I did not remove the torrents from the repo of course, just the links to them on that page.

The only real "negative" to this is that it means Old 3DS users will have to restore a NAND backup after they install Arm9LoaderHax, but realistically they should have to do this after the 2.1 CTR Transfer anyway, **The 2.1 CTR Transfer will totally wipe out tickets, and thus make them lose any digital games they had installed, as well as their save files, unless they somehow manage to soft install their games.** By making them restore their old NAND after installing Arm9LoaderHax, they will get their games and save files back, with no hassle. 

Preview the mostly new page here if you'd like.
https://cheatfreak47.github.io/Guide/installing-arm9loaderhax

If you'd rather stick with FBI Injection, I understand that too, but I figured it'd be worth exploring the alternative. `¯\_(ツ)_/¯`

### DIFF
--- a/_pages/DSiWare-Downgrade.md
+++ b/_pages/DSiWare-Downgrade.md
@@ -52,7 +52,6 @@ System Transfers can only be performed once a week.
 * The latest release of [3DSident](https://github.com/joel16/3DSident/releases/latest)
 * The latest release of [FBI](https://github.com/Steveice10/FBI/releases/latest)
 * The latest release of [dgTool](https://github.com/Plailect/dgTool/releases/latest)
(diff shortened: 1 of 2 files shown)


Plailect:
I'd rather use FBI injection, but I'll address those issues.

PR #701, opened 2016-11-14
Add trackers to the magnet links

DanTheMan827:
The magnet links didn't have any trackers so people using them may not find as many peers as if they would have used the torrent file.


Plailect:
DHT makes trackers on magnet links unnecessary, doesn't it?

DanTheMan827:
It can be used without trackers but DHT, PeX, and LSD may not find all of the peers that exist on the trackers.

I just did a test of the magnet links without peers and I got 33~ and as soon as I added the trackers it went up to 70~

PR #735, opened 2016-11-29
Update DSiWare-Downgrade.md

mtjohnson47:
Fixed ordering of steps and fixed numbers


Plailect:
The numbers don't actually matter in markdown.

PR #764, opened 2016-12-09
Possible typo

Ev1lbl0w:


Plailect:
"EmuNAND" is a more commonly used term. Not many people use RedNAND without a9lh of some kind.

Ev1lbl0w:
No, what I am talking about is that in the text it's written "EmuNAND/EmuNAND" I thought it should be "EmuNAND/RedNAND"

PR #799, opened 2016-12-22
Anti-Panic message for the softbrick

miltoncandelero:
While I was downgrading my 3ds, sysDowngrader failed and I landed on the Nintendo error message.

I panicked and I almost go with the *"boot holding A + R + L + Up on the D-Pad"* without trying to boot normally first.

Luckily while in panic I turned it off and on again. Turns out I just had a partial downgrade.

Still I was really close to updating my console due to sheer panic.


jaxrogers2:
This is what the line "If you get an error in the middle of installation, follow the instructions in the beginning for partial downgrades" means. It could be better clarified to mention include the Nintendo Error screen specifically, but otherwise it's already in the guide.

Plailect:
https://github.com/Plailect/Guide/commit/2d3d3ccf2a36a6a2ecda8d8f575af64070ea8836

No longer needed.

Issue #820, opened 2017-01-03
Suggestion regarding reporting issues


sora10pls:
Perhaps have a button somewhere at the top of each page in the guide that links people to https://github.com/Plailect/Guide/issues/new where they can report issues if they feel it's necessary

![image](https://cloud.githubusercontent.com/assets/17801814/21600567/6ff3ef42-d14c-11e6-883d-85a607cdcd74.png)

mariogamer2:
You also should have a .github folder with an ISSUE_TEMPLATE.md in it (default text in an issues).

mariogamer2:
#822

PR #822, opened 2017-01-03
Issue template (based on gogits/gogs one) and #820

mariogamer2:


mariogamer2:
Yeah I agree :P (#issuesspamming)

PR #834, opened 2017-01-05
Organisation for all the "what you need" sections

quantumcat1:
-All dot points for the "what you need" sections moved into files.xml, organised with xml structure
-"addfiles.html" included in the "what you need" section which uses functions.js to construct the dot points

Benefits:
-Edit a label/link pair in **one** place when it is used across multiple pages, for much easier maintenance
-Tidier pages so it is easier to see & edit the instructions (no longer have to scroll through lines and lines of magnet links)
-A lot less copy and pasting for you when creating multiple pages using some of the same files, and less possibility for mistakes
-I am going to write a Java program to download the files per page, and the structure is needed to find out which files are required (too hard to parse the html DOM)

This is my first pull request ever so I hope I haven't done anything wrong. I tried to put back all the config files I changed to make your site work on my github, hopefully I didn't miss any!


Plailect:
Gonna be honest, I don't really see the advantage to this. It seems significantly more complicated.

Can't you just parse the markdown files for links?

quantumcat1:
Yeah that's fine I can do that. When I started the html looked so complicated and I didn't know what markdown was.

quantumcat1:
Ill change my program to parse the markdown pages, and if successful can I later add some div marks to above and below each "what you need" section to help differentiate from other links on the pages? If it doesn't work out I can just keep my fork updated with changes and the program can run from there.

Plailect:
All you have to do is parse the markdown between "what you need" and "instructions".

Also, if I'm going to be honest, this is more of a script and something like Python seems more applicable.

PR #855, opened 2017-01-07
Clarified how to restore sysNAND

IvanFon:
A lot of people have been verifying that to restore the sysNAND they have to go to "SysNAND Restore (keep a9lh)". I feel that adding the specific menu item to the guide will help.


kevinchau321:
@IvanFon 
This clarification is a good idea, fix the merge request so it can be pulled in!

PR #870, opened 2017-01-08
Update Troubleshooting.md

rashevskyv:
Safe now


Plailect:
This will not always be the case.

PR #874, opened 2017-01-09
Update DSiWare-Downgrade-(App-Injection-and-Second-3DS).txt

rashevskyv:
Insert\reinsert target 3ds card


Plailect:
The first one is not consistent with the rest of the guide.

PR #878, opened 2017-01-09
Update Installing-arm9loaderhax.txt

rashevskyv:
Need update just before FBI inject. If someone with low firmware and a9lh try just to install FBI - he cant without update


CrystalMoogle:
Isn't that their fault for not following the guide..? You have to update prior to that point. It's not optional

PR #887, opened 2017-01-10
Change wording to emphasize that you NEED A BACKUP

ItsKipz:


Plailect:
This really doesn't get the point across, plus the backup doesn't have to be created at that step. Some people will already have one.

PR #890, opened 2017-01-10
Clarification on re-enabling wifi 

AnalogMan151:
Added extra clarification on how to re-enable wifi on 2DS and n3DS for people who are misunderstanding the troubleshooting step and leaving the AC charger plugged in.

### DIFF
--- a/_pages/en_US/2.1.0-ctrtransfer.txt
+++ b/_pages/en_US/2.1.0-ctrtransfer.txt
@@ -5,7 +5,7 @@ lang: en_US
 ref: 2.1.0-ctrtransfer
 ---
 
-If you downgrade to 2.1.0 on a 2DS or New 3DS and left Wireless Communication off, you can re-enable the wireless by removing the battery for several seconds then booting back up.
+If you downgrade to 2.1.0 on a 2DS or New 3DS and left Wireless Communication off, you can re-enable the wireless by removing the battery and unplugging the charger for several seconds then booting back up.
 {: .notice--info}
 
 If you've been using the New 3DS's microSD Management to transfer files to your SD card, this will no longer work on 2.1.0, make sure to have a microSD card reader on hand before proceeding.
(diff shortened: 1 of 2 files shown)


jkcgs:
The bug also affects to Old 3DS users? Because 2DS and New 3DS does not have the switch

AnalogMan151:
Correct, this is not an issue on o3DS because it has a physical switch. This is clarification just for 2DS and n3DS.

jkcgs:
Then the second commit is not valid, right?

AnalogMan151:
I'm not sure what you're referring to. I put my reasoning in the commits description. Model of 3DS has nothing to do with it.

jkcgs:
So my question is, do you actually need to remove the battery on Old 3DS while on 2.1.0 to turn on Wireless Communication?

AnalogMan151:
No, you don't. But that page is not specific to the o3DS, it's used in downgrading all models and so it's useful to have there. Would you prefer if I changed it to this: "If your wifi switch is off and you have a 2DS or n3DS, unplug the 3DS and remove the battery for several seconds to enable wifi

jkcgs:
With "If your wifi switch is off and you have a 2DS or n3DS", you are not talking about a physical switch? Because they don't have it.

AnalogMan151:
Correct. Glad we could sort that out.

jkcgs:
Great, thanks for this clarification of the clarification 😂

AnalogMan151:
If it makes it clearer for you, the reference with the wifi switch is because that's the error message the 2DS or n3DS says on-screen while on system version 2.1. That's the exact wording it uses because 2.1 assumes you have a physical switch due to the 2DS and n3DS not being released at the time. So if you see the message about your wifi switch being off, then those are the steps you follow.

jkcgs:
I got it now. Maybe you should clarify that you are refering to an error message, quoting it, or something like that.

PR #917, opened 2017-01-16
Update Get-Started-(New-3DS).txt

rashevskyv:


Plailect:
Definitely the first two columns.

PR #934, opened 2017-01-19
Update Region-Changing.txt

rashevskyv:
Why not? Guide is multilang now


Plailect:
make it "Choose a new language if needed"

PR #938, opened 2017-01-19
Update Installing-arm9loaderhax.txt

rashevskyv:
no chainloader yet


Plailect:
It's actually still called the chainloader (when buttons are held), just the menu isn't in stable.

PR #977, opened 2017-02-27
Stop DCIM from being deleted.

r3muxd:
Deleting DCIM will delete any photos you saved to SD. Kinda important. Doesn't fix i10n, and there's probably more than one part of the guide that tells you to delete it, so might need some improvement.


Plailect:
tbh, I forgot people actually take pictures with the 3DS

I'll make a proper fix soon

PR #995, opened 2017-03-23
Use X to to start GodMode9

Pitel:
Start button will be used as payload selector menu in Luma. Currenty does that in hourly builds.


Plailect:
I will change it when that's released.

PR #1027, opened 2017-04-16
include splash screens, freeshop link, and extra warning in home page

mmands994:


Plailect:
1. Definitely not including a link to piracy software in this
2. People can definitely update 99% of the time without being told?

PR #1030, opened 2017-04-17
fix the error in line 96

superboom:
"Select" actually launches Luma settings, "Start" is for the chainloader


CorrosiveCat:
I believe you got a bit confused as (star)t is used here to tell GodMode9 to reboot, (select) does not achieve this

PR #1065, opened 2017-05-08
Little fix

C0mm4nd:
Removed a duplicate sentence


Plailect:
fixed in latest commit manually

PR #1138, opened 2017-05-28
Temporary fix to Finalizing Setup and Updating to Boot9strap to (hopefully) avoid the banwave

r3muxd:
This shouldn't be permament, but this might need to stay up for the length of the banwave. (The other branches and languages need to be updated, but I don't know how.) All credits go to sirocyl on the Discord.

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,0 @@
-3ds.guide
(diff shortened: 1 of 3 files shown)


Plailect:
This seems unnecessary as the cause of the bans was pretty much pinned down to SpotPass and the friend's list thing, which the guide instructs users to disable.

CNAME
```diff
@@ -1 +0,0 @@
-3ds.guide
```
Plailect: uh, why?

r3muxd:
Sorry. I was trying to test it out.

PR #1175, opened 2017-06-03
Remove jksm

rashevskyv:
Cannot use it without hack.


Plailect:
Userland entrypoints exist on 11.4.0. I will link to the alternatives page once that is updated

PR #1198, opened 2017-06-07
updating Guide

architdate:


architdate:
My mom was told to not touch my PC.

PR #1206, opened 2017-06-10
Edit get started page

Multipixelone:
This just makes it easier to understand. Kinda confusing mentioning 3DS and 2DS but not N3DS.


Plailect:
New 3DS is included in "3DS"

PR #1214, opened 2017-06-11
U cannot use soundhax on fw without browser

rashevskyv:
U cannot use soundhax on fw without browser couse there is no otherapp for browser version lower than 7 and if u got console with fw bigger than 9.0.0 without browser it is clear that console was updated with cartrige. So u cannot use soundhax in both cases.


Plailect:
This may be true, but to say that a browser is required for soundhax to work would be misleading (even if it is impossible to get on a firmware with latest soundhax but no browser). If that were possible, you could just use the -7 otherapp and it'd work just fine. I'll leave it as is because it's more technically accurate even if it is not a possible situation.

PR #1226, opened 2017-06-17
New 2DS

rashevskyv:


sora10pls:
This would also have to accommodate for the [Old 3DS Get Started page](https://3ds.guide/get-started-(old-3ds)), as it never said "or 2DS".

PR #1241, opened 2017-07-10
Update 404.txt

Cecil107:
Translated to czech language


Plailect:
This is not the place to do that.

http://translate.3ds.guide

PR #1252, opened 2017-07-22
delete nand backup section, feature now disabled on latest godmode9

mmands994:


Plailect:
It's done with the scripts section...

PR #1288, opened 2017-08-14
Update installing-boot9strap-(dsiware-game-injection).txt

Ninjistix:
clear up some confusion, (for myself and others)


Plailect:
This is not the case for all games. Only some use `savedata.bin`.

Ninjistix:
Oh okay, I was confused when the guide said all contents when it was just
the savedata.bin

PR #1295, opened 2017-08-17
Add single-system NTRBootHax

ItsKipz:
This is a method that allows you to use NTRBoot using only one 3ds and an r4i, although it has not been fully tested on N3DS.


Plailect:
Implemented manually

PR #1314, opened 2017-08-25
Update finalizing-setup.txt

Samplasion:
I fixed a writing error


Plailect:
This should be addressed on the translation site, not here.

PR #1382, opened 2017-10-07
Update the Get Started page to better describe the Friend List 

thejsa:
Adds an icon and uses Nintendo's preferred spelling of "Friend List" (as opposed to "Friend's List" - which suggests a list possessed by a friend)


Plailect:
I like the idea but the wording isn't the best and that notice appears in quite a few places so it has to be changed for all of them. I'll do this myself when I have time.

Plailect:
Also getting kicked out of the menu indicates that it is already disabled (never set up in the first place), not necessarily that they have already been banned.

PR #1391, opened 2017-10-11
note that ctrtransfer needs to be of the region you want to change to

### LINKED ISSUE #1389: Region changing guide
"The 11.5.0 CTRTransfer image for your device and region 
(if your device is not from one of these regions, just pick one that matches your device type):"

This sounds like they are supposed to pick the region they are currently on. The second part makes that more clear (for example if you're trying to switch to Taiwan region, surely picking a random region isn't going to help you achieve that, so it HAS to mean pick the region you're on already). I don't see anywhere in the rest of the page that would help them switch regions though so I'm assuming they aren't actually supposed to pick the same region that they already have.

figgyc:
fix #1389


Plailect:
Why did this edit a bunch of other stuff?

figgyc:
I'm not sure to be honest probably just my editor acting up, anyway i'm fairly sure the lines are the same

PR #1402, opened 2017-10-23
Checkpoint instead jksm

rashevskyv:
Checkpoint instead jksm


fryfrog:
Why not show both: "use the new [Checkpoint] or the old [JKSM]" maybe?

figgyc:
Well theres no real point in using jksm anymore

fryfrog:
¯\_(ツ)_/¯

PR #1405, opened 2017-10-24
Add Ace3ds X (comes w/ ntrboot and switches between it and DS cart)

fryfrog:
This one links to the team page, leaving finding it to the reader. It is also sold at http://3ds-flashcard.com/home/65-ace3ds-x-supports-3dsds-mode.html so that URL could be used instead. Or if nds-card starts selling it, link there.


Plailect:
There's no reason to include this in the guide because it does not require flashing. Users of this cart already get instructions to use it, or could just begin at https://3ds.guide/installing-boot9strap-(ntrboot)

PR #1408, opened 2017-10-27
Update ntrboot links for ntrteam org

figgyc:
They moved the flashers to a single repo.
~~Note that ds_ntrboot_flasher's release page still uses the old filenames; this will need to be changed at a later date.~~ That was changed already


Plailect:
fixed without realizing this PR already existed

PR #1421, opened 2017-11-02
Revert all recent changes to a9lh-to-b9s, and also a minor modification

jason0597:
I reverted a9lh-to-b9s.txt to how it was on https://github.com/Plailect/Guide/commit/32680e2d4d5d5226748517e47f15f4f15b99513a
Admittedly I did not think about old versions of a9lh that aren't able to boot `arm9loaderhax_si.bin` payloads. Perhaps this can be looked at in the future if/when SafeB9SInstaller has built-in screen init. For the time being, the guide will have to stay as it is.

I also modified it a bit because it said "some versions of luma won't autoboot the payload starting with start_" even though the guide guarantees that the user will be on luma 7.0.5

I will note that there is a problem as it is right now: A lot of people update from ancient versions of luma to luma 7.0.5, and it asks them to reconfigure luma. People will freak out, and they will come to the discord channel asking for help "halp i hold start on bootup but i go into 'luma configuration menu' please help". 
I did not do any modifications to fix this.

I apologize for any inconvenience that the issue I opened may have caused.

### DIFF
--- a/_pages/en_US/a9lh-to-b9s.txt
+++ b/_pages/en_US/a9lh-to-b9s.txt
@@ -81,9 +81,6 @@ For all steps in this section, overwrite any existing files on your SD card.
 #### Section II - Installing boot9strap
 
 1. Boot your device while holding (Start) to launch the Luma3DS chainloader menu
-  + Some versions of Luma3DS will just directly start whichever payload begins with `start_`
-  + If your version does this, just proceed with the instructions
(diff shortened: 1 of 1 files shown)


trbocode:
Why reverting Luma 3DS updater?

jason0597:
Addressed in https://github.com/Plailect/Guide/pull/1422

PR #1427, opened 2017-11-07
Update cleanup_sd_card.gm9

rashevskyv:


figgyc:
Has literally anything changed? I can't see it. Was the commit an accident?

figgyc:
nvm i see it, might be a better idea to merge this into the same PR as #1426 to avoid confusion

PR #1432, opened 2017-11-11
tell users what to do if they get hbl error

jason0597:
Yes they may have been told in previous parts of the guide to download the latest one and overwrite the old one, but really, if they are at this point where they see an error they most likely weren't paying attention and so telling them to redownload it isn't bad (since if they see this screen then they most likely __did not__ download the latest one)


Plailect:
This is unnecessary because:
1. They should read instructions
2. They would have already downloaded the file if not copied it

Also this PR links directly to a specific release rather than using the `/releases/latest` link

PR #1448, opened 2017-11-19
Tell them they need free space for NAND backup

jason0597:
To prevent incidents like this one:

img


jason0597:
When we get if statements in GM9, this will be reverted

PR #1467, opened 2017-12-20
Update godmode9-usage.txt

TheSwitchGuy:
Make it so people know the danger of remove the NNID with the script


Plailect:
This isn't really necessary

PR #1512, opened 2018-03-17
Open otherapp payload link in new tab

brendanmullan:
Otherapp payload link opened in same window, updated to open in new  tab as per the other links.


Plailect:
Pull requests should only modify en_US pages. The others will be updated through the translation system.

PR #1544, opened 2018-06-10
adds step to ensure user remounts SD card before moving to next section

test-as-other-user:
Encountered a file-not-found error at step 2 below, because SD card needed to be remounted.

#### Section VIII - Cleanup SD Card
  1.  Press (Home) to bring up the action menu
  2. Select “Scripts…”
  ...


figgyc:
The SD is supposed to remount itself and usually it does so you should probably say that you should check if the SD card has reappeared and remount it if it hasnt

Plailect:
Yep, this is exactly as figgyc said.

PR #1546, opened 2018-06-12
comprehensive format+sector check instructions

eip618:
I noticed that this page includes only instructions for running h2test as if it's on a new card. These instructions include formatting help, which should clear up any sd card issues that may be present.


PixelSergey:
`Move, to your computer, as much of your own data as you can.`
Could this not just be `Back up the contents of your SD card to your PC`? Just saying the former would get *some* user to not back up their `Nintendo 3DS` folder or the like

Plailect:
I don't really think this makes sense as a location for formatting instructions (it is just to check the SD for issues prior to the guide), plus the language used in the PR is not consistent with that of the rest of the guide.

PR #1548, opened 2018-06-13
Fixing little typos, missing spaces, and other minor issues in PL translation.

Caier:
All of these changes were also done in crowdin almost a week ago, but they weren't added.


Plailect:
Translation stuff should never be manually added.

PR #1552, opened 2018-06-20
11.7 on all languages and some fixes

lordfriky:
For some reason, I can't fully edit some Hebrew lines in Atom, so I use the GitHub web for editing those later. And the fixes are some things that I found while I was editing all the languages.


Mike15678:
@lordfriky I’m pretty sure this is already being done through Crowdin.

lordfriky:
Oh, doesn't matter, at least Plailect can see the fixes I made

Plailect:
All translations are managed via Crowdin, this is not the place to make changes.

PR #1607, opened 2018-11-23
edit frogminer steps to include new frogtool

eip618:
frogminer has been updated to bundle ctcert, the srl and also be able to patch ds dlp on-device, meaning jenkins is no longer needed. this update reflects that change


eip618:
if you merge this, merge patch 2 because i couldnt commit directly to patch 1

zacchi4k:
Actually, the new frogtools fails on my console, and zoogie and I are trying to figure that out before it can become official

eip618:
sorry

PR #1609, opened 2018-11-23
steel diver is not dsiware

eip618:
edited to change that, also added a note about spamming bfm refreshes, edited introduction to remove reference to "several services" (?)


eip618:
sorry

PR #1614, opened 2018-11-27
Update installing-boot9strap-(frogminer).txt

TheGinGear:
Can't launch steelhax after cfw is installed. Use Rosalina menu to access frogtool instead to restore clean ds download play


eip618:
i've put the content of this in #1615 as well

PR #1615, opened 2018-11-28
fresh magnet + section 6 fix

eip618:
i have added a new magnet link because of an issue we identified and hopefully have resolved
also fixed section 6 (thanks gingear) because steelhax doesnt work after installing b9s and luma


eip618:
closing because now i've removed the magnet, it's exactly the same as gingear's pr
:blobshrug:

PR #1621, opened 2018-12-04
update to include twl test

eip618:
some systems, usually due to prior uninstallation of cfw, have a broken twl firm and cannot use frogminer. it is pointless for such users to proceed through these steps only to find they can't do it. this can be detected from stock very easily and this pr includes steps for such a test

### DIFF
--- a/_pages/en_US/installing-boot9strap-(frogminer).txt
+++ b/_pages/en_US/installing-boot9strap-(frogminer).txt
@@ -37,10 +37,10 @@ This process will overwrite your game's save file!
 
 #### Section I - Prep Work
 
-1. Go to System Settings, then "Internet Settings", then "Nintendo DS Connections".
-1. Press "OK".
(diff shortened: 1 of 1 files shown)


Plailect:
It feels unnecessary for all users to test this,nit should probably just be for previous users of CFW.

Sonlen1414:
So having someone take a minute or two to test if a method will work before wasting potentially hours later on is unnecessary? That seems like some very flawed logic to me, especially since you can't prove a pre-owned console never had CFW before.

eip618:
just had this: 
https://imgur.com/a/lOBOZhQ
as context for the issue. this isn't the first time we've had something like this pop up

Plailect:
Please review the formatting changes I mentioned on GitHub. For examples, look at how links to troubleshooting pages are typically handled. Also subtext is not a full sentence and should not have any periods in it, please rephrase to remove them.

PR #1622, opened 2018-12-04
11.9

zacchi4k:
Firmware 11.9.0 has been released, and it's exactly the same as 11.8.0 when it comes to modding


zacchi4k:
Note that the otherapp.bin page on smea's website hasn't been updated to include yet, but the otherapp payload for 11.8 (and 11.7) works fine, I wasn't sure whenever to add a disclaimer about that. The site should soon get updated tho

PR #1626, opened 2018-12-14
Update instructions for Frogtool 2.1

zacchi4k:
Frogtool 2.1 requires a separate _frogcert.bin_ file on the SD card root (which contains the Flipnote SRL and ctcert .bin) downloaded via torrent, meanwhile Frogtool itself can be hosted on GitHub again. Note that the rss.xml hasn't been updated to include frogcert.bin (the old Frogtool magnet can be removed)


Plailect:
I'll likely have to change this magnet link to get it to play nice with the torrent RSS, but otherwise looks good.

PR #1635, opened 2019-01-29
Cia file notice for hbl wrapper

MechanicalDragon0687:


Plailect:
This notice was in the past only used to tell the user which files to download when there are multiple. In this case, there's only one file on the release. Why is this needed?

MechanicalDragon0687:
For some reason people see source options and downoad them since all others seem to indicate which file...

MechanicalDragon0687:
What about a note to not download source files then since that is global for all of them?

Plailect:
I don't think this is really needed and this has been sitting for a while. Closing unless further discussion of this comes up.

PR #1639, opened 2019-02-08
Move ds internet restore from its awkward spot in finalizing to the end of the fredtool guide itself; other minor things

aspargas2:


aspargas2:
added removal of frogcert link

PR #1651, opened 2019-02-27
Added boot9.bin to Line 149

UmbraSage:
Hopefully clears up some slight confusion on what to delete and what to keep

~~and small spacing on Line 146~~


Plailect:
There's no reason to delete this, it takes up practically no space and keeping a copy on the SD may be useful if they lose the other one.

UmbraSage:
I can understand the space size. 
Trying to quote asp (in a weird summarized way), 'that essentials can work as a backup. It's just a hard way of restoring the 3DS.'

PR #1652, opened 2019-02-28
add note about seeding

eip618:
have seen *a lot* of confusion about this and thought a note may help


Plailect:
I think explaining how torrents work is a bit outside the scope of the guide

PR #1656, opened 2019-03-18
comprehensive uninstall instructions

eip618:
add extra depth to uninstall instructions incl a safety check for brickability
also add deleting cfw apps to avoid permacrashing data management


MechanicalDragon0687:
` should be manually removed before starting this process using the Data Management`
i would recommend 
` should be manually removed before starting this process by using the Data Management`

PR #1657, opened 2019-03-18
Add short explanations for components of the Seedminer finalization step 

Redmega:
Adds a short description of what DSP1 and `ctr-no-timeoffset` are responsible for in the *What You Need* section.

Thanks to users `ahrigatø` and `bungiefan(what does the guide say?)` from Discord, who explained what `DSP1` and `ctr-no-timeoffset` were for.


Plailect:
The download space is not a good place for those. Notice that there is a section almost right above that where the purpose of other homebrew applications is explained.

Redmega:
Good point @Plailect. I didn't put these there because they're not mentioned there before, but I'll unify them now.

Plailect:
The purpose of DSP1 is so that homebrew applications have sound, nothing to do with CIA formatted games.

Redmega:
Ah, that was a misunderstanding of mine. Should be good to go now.

Plailect:
Alright, now just fix the capitalization and I'll merge (other descriptions do not begin with any capitalization).

Redmega:
Whoops

PR #1660, opened 2019-03-19
Remove the need for torrenting anything for fixing TWL after installing b9s

aspargas2:


aspargas2:
the twl cia torrents could also be removed from the rss feed after this is merged, however one does that

PR #1685, opened 2019-05-28
Update for 11.10.0

ihaveamac:
11.10.0-43 was released recently, but does not include any important
changes.


Valeri0p:
EUR home menu was updated, so we need a new otherapp payload for the "Seedminer + Steelhax + Frogtool" method, since Steelhax doesn't seem to work.
Or you can buy any dsiware to bypass the problem anyways...

PR #1691, opened 2019-06-29
use finalizing setup

MechanicalDragon0687:
there was duplicated data in the a9lh to b9s guide, which is convered in the finalizing setup, instead i pointed users to finalizing setup

in addition, this necessary data was missing from the b9s1.0 to b9s 1.3 update guide, i also point that at finalizing setup

As normal, this is en_us only


MechanicalDragon0687:
resolved conflicts and merged

PR #1695, opened 2019-07-09
change steelhax to rehost of zip

MechanicalDragon0687:
Asked VegaRoXas for permission to rehost as zip with only steelhax folder. 
permission granted. 
modification should make it less error prone for users copying files


MechanicalDragon0687:
Permission email screenahots https://github.com/MechanicalDragon0687/vegaroxas.github.io/blob/master/Screenshot_20190709-141721.png
https://github.com/MechanicalDragon0687/vegaroxas.github.io/blob/master/Screenshot_20190709-141735.png
https://github.com/MechanicalDragon0687/vegaroxas.github.io/blob/master/Screenshot_20190709-141746.png

PR #1697, opened 2019-07-18
Updated NTRboot guide with info on R4i Gold 3DS plus

radostin04:
Made the NTRBoot guide more clear on the R4i Gold 3DS plus being preflashed and how to switch to the ntrboot mode.


Plailect:
This doesn't really seem to belong here? It provides weirdly preferential treatment to reference this cart and only this cart outside of the main table (and this is info already elsewhere on the page...)

PR #1725, opened 2020-01-16
Add note about how to get essential.exefs if it is missing

JasonCubed132:
I've been informed that this should be created by Godmode9's backup prompt when you first run it, but I've seen several users ask about how to get this as they don't have it so I've added it in.


Plailect:
I think this should be reported as a bug upstream to GM9, not addressed here. Users will absolutely ignore that instruction as added if they don't believe it to be absolutely required (and too much effort).

JasonCubed132:
Closing as https://github.com/nh-server/Kurisu/pull/733 has been suggested.

PR #1730, opened 2020-03-25
Add homebrew launcher to finalizing-setup.txt

missionfloyd:
ntrboot users may not have already copied the homebrew launcher, causing an exception when trying to launch it.


ItsPizzaTime1501:
in my opinion this will confuse users a lot more because every guide links to the finalizing setup and most of them have the user put boot.3dsx on the sd root, adding that to the finalizing-setup would probably just cause more confusion

missionfloyd:
Apparently it's included with luma now.

PR #1731, opened 2020-03-26
Wrong version and missing translation

Tom60chat:
Version Table:
11.4.0 / 11.9.0 => 11.4.0 / 11.11.0
11.10.0 / 11.11.0 => 11.12.0 / 11.13.0
Update to latest firmware or use an "All Versions" Method => Mettez à jour vers le dernier firmware ou utiliser la méthode "Toutes les versions"


Plailect:
Translations should not be manually edited as they will be overwritten by the automated crowdin merges. If this has already been fixed on crowdin, I will need to initiate a merge.

PR #1733, opened 2020-04-10
Update get-started.txt

Oklahomawhore:
fix the version table not correct problem.


Plailect:
Translations should be modified on Crowdin, not edited manually. They would be overwritten if you did it this way.

PR #1754, opened 2020-04-27
various changes to streamline text and reduce redundancy

lilyuwuu:
Various changes have been made in this PR to aid in the transition to USM as the primary Seedminer method.

- Wording has been slightly changed when referring to other seedminer exploits, as they have been moved to the legacy methods page.
- The DS Internet Settings test has been replaced with a DSiWare management test, as the the latter is required for USM.
- "Continue to unSAFE_MODE" has been changed to "Continue to Installing boot9strap (USM)", in line with other pages on 3ds.hacks.guide.
- DS Internet and DS Download Play tests have been added to the Legacy Methods page, as well as a link to TWLFix-3DS should they be broken.
- Homebrew Launcher (Pichaxx) now links to Installing boot9strap (Frogtool), as Pichaxx + DSiWare Dumper + Fredtool is more of an edge case.

### DIFF
--- a/_pages/en_US/homebrew-launcher-(pichaxx).txt
+++ b/_pages/en_US/homebrew-launcher-(pichaxx).txt
@@ -50,5 +50,5 @@ If you already have Pokemon Picross, this process will overwrite your game's sav
 
 ___
 
-### Continue to [DSiWare Dumper](dsidumper)
+### Continue to [Installing boot9strap (Frogtool)](installing-boot9strap-(frogtool))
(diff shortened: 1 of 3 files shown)


GlaZedBelmont:
Looks fine to me too

though I'd personally reword "There are several different methods that build on the Seedminer exploit"
to 
"There are several different methods built around the Seedminer exploit" / "There are several different methods that are built around the Seedminer exploit"

PR #1759, opened 2020-04-28
finalizing-setup: use Luma3DS NTP sync

ihaveamac:
This replaces ctr-no-timeoffset and (in most cases) setting RTC via
GodMode9. This currently doesn't have a method to set the offset if NTP
sync fails though, since Luma3DS only sets the offset to 0 if it
succeeds.


aspargas2:
One possible solution would be to add a sub-bullet that if it fails, the user will need to manually correct the time later, but that would never actually set the offset to 0 because the user won't enter the current time perfectly.

PR #1763, opened 2020-05-04
Formatting SD guide (Mac)

ihaveamac:
Unfinished, needs steps for Yosemite/10.10 and before.


ihaveamac:
Disk Utility doesn't allow empty volume labels so I re-added 'Enter any label for "Name" except "3DS"' for now.

PR #1767, opened 2020-05-26
Add issue and pull request templates

FrozenChen:
This will help users format their issuers and collaborators give context to their contribution


NightYoshi370:
imo, I don't think the "Hacking Support" issue should be made. That should just be handled via the Nintendo Homebrew Discord server. The average joe looking at this would more likely have a Discord account rather than a GitHub account.

PR #1775, opened 2020-07-21
change bruteforcemovable links to alternatives

aspargas2:
This should be merged at 10 pm UTC when bruteforcemovable is scheduled to have downtime, then reverted once the downtime is over.


aspargas2:
Make that 10 pm UTC tomorrow.

aspargas2:
Make that... whenever this downtime actually happens.

aspargas2:
Which is apparently never.

aspargas2:
Will leave this branch here though, to make things slightly easier if it ever goes down in the future for whatever reason.

PR #1782, opened 2020-08-25
JS Dynamic Pages

NightYoshi370:
Replaces the table for JS users.
Keep in mind that this doesn't include the browser version soundhax checker. That is because this is the initial commit and as such, is simply a reinterpretation of the table. Just like the table doesn't accomodate the soundhax checker, neither does this.

Initial code (hardcoded) by me.
Revised class code by me.
Sorting code (for the dropdown) by Samplasion.


NightYoshi370:
I pushed a change that adds a checkbox to the BrowserHax 2020 page.

When you load the page, it will look like this
image

Notice how by default it is on.
Now what happens when you tick the checkmark off? It would change to this:
image

It should help out, but helpees are *interesting* so I didn't remove the notice from the main text.
Also, text appears different because I use a different CSS framework (my own) for testing and have not imported the Minimal Mistakes one (imo, minimal mistakes is extremely ugly but I digress). Functionality is no different though.

PR #1783, opened 2020-08-31
Add filler data to DO_NOT_LAUNCH.firm

romhacks:
This is a small change to add some text to assets/DO_NOT_LAUNCH.firm, as it seems that some browsers have trouble downloading the 0 byte file.


romhacks:
It seems that this was already added, probably part of the browserhax addition.

eip618:
hi, yes, this was resolved in some recent changes relating to browserhax. the original do not launch firm did actually have some placeholder data, it seems the one which made its way here did not. thanks for letting us know

PR #1789, opened 2020-09-08
Remove DO_NOT_LAUNCH.firm, switch to GodMode9 v1.9.2pre1

aspargas2:
I would like a few more people to agree with me on this course of action (and someone to look over the actual PR to make sure I didn't do anything silly) before this is merged.


lilyuwuu (approved):
Actual content of PR seems fine; no errors that I could see.

While `prerelease` is the correct term, it might seem a bit scary without context ("why do I have to use a pre-release, won't that be more buggy?" / "um, the release notes say that this is beta, should I still use it?"). 

Otherwise, I support the change

aspargas2:
I don't really see a way around people getting worried over the release notes, but do have any ideas for a better word than `prerelease`?

NightYoshi370:
Why not try "preview"?

lilyuwuu:
Could call it "The v1.9.2pre1 release" (like when the guide switched to 1.7.1) since we're going to have to switch back to latest once GM9 gets an update anyway

aspargas2:
I like lily's suggestion. To me, "preview" sounds like it is incomplete, perhaps less complete than the full release before it.

PR #1791, opened 2020-09-09
Replace the recommendation of people with current access to HBL to a different Slottool Page and add notices and fix outdated pages.

### LINKED ISSUE #893: Eshop error 005-5602 on region changed 2DS
Hi, I have a Japanese 2DS, I followed Plailect's guide exactly to install A9LH and change region to US. 
After all done, I create a new NID on new region, and then I access eshop, I get the error 005-5602.
Please tell me how to fix this? Thank you very much!

### LINKED ISSUE #1778: Region Changing (in theory) will always break eShop, guide says otherwise.
Allow me to explain a bit, since I think it is relevant.

Basically, I recently did some testing on three consoles, under an EmuNand (for testing, never actually practically use it) I was able to test region changing, and on all three consoles, they could not access the eShop after region changing.

By going a bit further, and with the help of some people in NH Discord, I was able to do the SSL Patch stuff and take a look as to where it fails during loading the eShop.

For reference, on a stock console without region changes, if the region in system settings is changed, two ninja/eCommerce things are sent, server notices region has changed, and tells the system to load the changed region prompt. Go through it, then the server does IdentityAuthenticationSOAP, which will return internal code 0 and restart the eShop.

On a region changed console, however, the first half remains similar, however, the difference is the second half, where, when the server does IdentityAuthenticationSOAP, it returns the following internally:

"IAS - Region Not Match"
Error Code: 958

Which basically tells the system to throw error 005-5958, aka one of the region changed error codes.

The reason I bring this issue up is that, in theory, this call is always made the moment the system region changes (i.e through region changing) and the eShop is launched, this effectively makes it to where any console region changed that has accessed the eShop before will lose the ability to access the eShop on the region changed region. However, the guide mostly says otherwise.

This also most likely means that if the person does have a working eShop after region changing they **cannot** change the region in system settings or they will most likely lose the eShop, that being said, I cannot physically prove that.

The point I'm trying to make is that the guide in regards to that section is entirely outdated, and, at the very minimum, should be a blanket statement stating something along the lines of "Region Changing, in the majority of cases, will make you lose access to the eShop"

Systems Tested:

(SSL Patched) - Nintendo 3ds XL - SysNand 11.13.0-45U - EmuNand 11.13.0-45E

Nintendo 2ds/New Nintendo 2ds XL - Same as Nintendo 3ds XL (minus SSL Patch)

TheMasterOfMike:
Basically this is the following:

- Rename HBL-USM to BH-Slottool (Mainly for consistency with the next point)
- Create HBL-Slottool Page
- Replace the recommendation on BH-Slottool if users have existing HBL entrypoint to HBL-Slottool (rather than Frogtool like before)
- Add a note to the Steelhax page saying that the method is no longer working (in reference to the main injector having been taken down and having it point to the page being outdated)
- Alter the Steelhax and PicHaxx pages so they link to both HBL-Slottool, or Frogtool if they do not have working shoulder buttons
- Alter the references on the original Installing Boot9Strap (unSAFE_Mode) page
- Alter some wording and mention HBL-Slottool and Frogtool on Legacy-Methods
- Heavily change the text used for region changing
(Fixes #893 and Fixes #1778 due to the above)
- Fix Site Navigation to make it both consistent across both Popular and All, to add HBL-Slottool, and rename a couple of pages so that they line up with the changes made above.


lilyuwuu:
My thoughts on this PR:
> Rename HBL-USM to BH-Slottool (Mainly for consistency with the next point)

See my comment on the next point.
> Create HBL-Slottool Page

I am against this change. People following the guide *normally* will have very little need to ever go to this page, as browser will rarely end up broken (I've yet to see a case of a broken browser that is also on 11.13). While PicHaxx -> slotTool would technically be faster for people following video guides and only getting as far as pichaxx, it would be at the expense of confusing any non-video guide followers that are unfortunate enough to need to follow a legacy method. We also have a page off the guide if PicHaxx -> slotTool needs to be followed (that being said, it's my own site, so a bit of a conflict of interest here).

I believe the costs of separating the page into two (confusion) outweigh the benefits (time saved for edge cases).

> Replace the recommendation on BH-Slottool if users have existing HBL entrypoint to HBL-Slottool (rather than Frogtool like before)

See comment above.

> Add a note to the Steelhax page saying that the method is no longer working (in reference to the main injector having been taken down and having it point to the page being outdated)

This change is fine, I guess, but I think it would be good to delete the instructions of the page (not the page itself -- don't want to create broken links) and replace it with this Alter the Steelhax and PicHaxx pages so they link to both HBL-Slottool, or Frogtool if they do not have working shoulder buttons
    Alter the references on the original Installing Boot9Strap (unSAFE_Mode) page
    Alter some wording and mention HBL-Slottool and Frogtool on Legacy-Methods
    Heavily change the text used for region changing
    (Fixes #893 due to the above)
    Fix Site Navigation to make it both consistent across both Popular and All, to add HBL-Slottool, and rename a couple of pages so that they line up with the changes made above.

_pages/en_US/homebrew-launcher-(steelhax).txt
```diff
@@ -6,11 +6,14 @@ title: "Homebrew Launcher (Steelhax)"
...
+This method of using Seedminer for further exploitation uses your `movable.sed` file to gain access to the Homebrew Launcher. This method requires you to already own (or download) the free "Steel Diver: Sub Wars" game from the eShop.
 
-If your device already has access to the Homebrew Launcher (whether through Steelhax or another exploit), you can skip to [DSiWare Dumper](dsidumper)).
+If your device already has access to the Homebrew Launcher (whether through Steelhax or another exploit that is not BrowserHax 2020), you can skip to [HBL-Slottool](installing-boot9strap(HBL-Slottool)).
 {: .notice--success}
 
+This is an outdated method and is no longer working, these steps only remain for users who either already have the files to preform the exploit or already have access to the Homebrew Launcher, for anyone who does not have the files to preform this method, you should follow [BrowserHax 2020](homebrew-launcher-(browserhax-2020))
```
lilyuwuu, line 14: This is a run-on sentence.

_pages/en_US/homebrew-launcher-(steelhax).txt
```diff
@@ -59,5 +62,8 @@ If you already have Steel Diver: Sub Wars, this process will overwrite your game
 
 ___
 
-### Continue to [DSiWare Dumper](dsidumper)
+### If you have working shoulder buttons, continue to [Installing boot9strap (HBL-Slottool)](installing-boot9strap-(HBL-Slottool))
+{: .notice--primary}
+
+### If you do not have working shoulder buttons, continue to [Installing boot9strap (Frogtool)](installing-boot9strap-(frogtool))
```
lilyuwuu, line 68: If this method is outdated, why continue to maintain it?

_pages/en_US/installing-boot9strap-(BH-Slottool).txt
```diff
@@ -8,11 +8,11 @@ title: "Installing boot9strap (HBL-USM)"
 
 In order to exploit the SAFE_MODE firmware of our system, we need to inject an exploited WiFi profile.
 
-As we already have Homebrew access, we can use slotTool to do this.
+As we already have Homebrew access using BrowserHax 2020, we can use slotTool to do this.
 
 This specific method is intended to be used with Homebrew Launcher access through [BrowserHax 2020](homebrew-launcher-(browserhax-2020)).
 
-If you have Homebrew Launcher access through another entry point, you should use [Frogtool](installing-boot9strap-(frogtool)).
+If you have Homebrew Launcher access through another entry point, you should use [HBL-Slottool](installing-boot9strap-(HBL-Slottool)).
```
lilyuwuu, line 15: People shouldn't end up at this page unless they're following Browserhax 2020. Adding additional points only increases chance of confusion.

_pages/en_US/installing-boot9strap-(HBL-Slottool).txt
```diff
@@ -0,0 +1,85 @@
...
+
+### Required Reading
+
+In order to exploit the SAFE_MODE firmware of our system, we need to inject an exploited WiFi profile.
+
+As we already have Homebrew access using an older entrypoint, we can use slotTool to do this.
+
+This specific method is intended to be used with older Homebrew Launcher access methods, such as PicHaxx and Steelhax.
+
+If you do not have Homebrew Launcher access currently, you should use [BrowserHax 2020](homebrew-launcher-(browserhax-2020)).
```
lilyuwuu, line 15: A bit weirdly worded

_pages/en_US/installing-boot9strap-(HBL-Slottool).txt
```diff
@@ -0,0 +1,85 @@
...
+
+If you currently have Homebrew Launcher access using BrowserHax 2020, you should follow [BH-Slottool](installing-boot9strap-(BH-Slottool))
```
lilyuwuu, line 17: This sentence does not have a period at the end

_pages/en_US/installing-boot9strap-(HBL-Slottool).txt
```diff
@@ -0,0 +1,85 @@
...
+### What You Need
+
+* The latest release archive for [Luma3DS](https://github.com/AuroraWright/Luma3DS/releases/latest)
+* The latest release archive for [unSAFE_MODE](https://github.com/zoogie/unSAFE_MODE/releases/latest)
+
+#### Section I - Prep Work
+
+1. Copy `boot.firm` from the Luma3DS `.zip` to the root of your SD card, replacing any existing files
+1. Copy `usm.bin` from the unSAFE_MODE `.zip` to the root of your SD card
+1. Copy the contents of the `slottool` folder from the unSAFE_MODE `.zip` into the `3ds` folder on your SD Card
```
lilyuwuu, line 35: The second T in `slotTool` should be capitalized

_pages/en_US/installing-boot9strap-(usm).txt
```diff
@@ -16,8 +16,11 @@ Once the WiFi profile has been injected, we will use SAFE_MODE, which is a recov
 
 These instructions work on USA, Europe, Japan, and Korea region consoles as indicated by the letters U, E, J, or K after the system version.
 
-If your (Right/Left Shoulder), (D-Pad Up) or (A) buttons do not work, you will need to use a [Legacy Method](legacy-methods). For assistance with this matter, join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp) and ask, in English, for help.
-{: .notice--danger}
+If you have an existing Homebrew Entrypoint using BrowserHax 2020, you should follow [BH-Slottool](installing-boot9strap-(BH-Slottool))
+{: .notice--warning}
+
+If you have an existing Homebrew Entrypoint using something other than BrowserHax 2020, you should follow [HBL-Slottool](installing-boot9strap-(HBL-Slottool))
+{: .notice--warning}
```
lilyuwuu, line 23: People would only ever get linked to this page through the legacy method, so putting these warnings here would only cause confusion

_pages/en_US/legacy-methods.txt
```diff
@@ -14,11 +14,20 @@ However, "legacy" or outdated methods are kept here for various purposes. They w
...
+The following exploits make use of one of three DS-mode applications or DS-mode functions built into the 3DS: DS Internet Settings, DS Download Play, and DSiWare Data Management.
+
+If DS Internet Settings, DS Download Play, and DSiWare Data Management fail to work, you will have to repair them with [TWLFix-3DS](https://github.com/MechanicalDragon0687/TWLFix-3DS/releases/) using a homebrew entrypoint, such as BrowserHax 2020 or PicHaxx.
+
+#### DSiWare Data Management Test (used for unSAFE_MODE and Fredtool)
 
-If both DS Internet Settings and DS Download Play fail to work, you will have to repair them with [TWLFix-3DS](https://github.com/MechanicalDragon0687/TWLFix-3DS/releases/) using a homebrew entrypoint, such as Pichaxx.
+1. Go to System Settings, then "DSiWare"
+  + If the system does not crash to the Home Menu, the test was successful
+  + If the system crashes to the Home Menu, the test has failed
```
lilyuwuu, line 30: This test is actually irrelevant because apparently it'll only crash once you put a dsiware in the folder.

_pages/en_US/legacy-methods.txt
```diff
@@ -38,9 +47,11 @@ If both DS Internet Settings and DS Download Play fail to work, you will have to
...
+1. [unSAFE_MODE](installing-boot9strap-(usm)): Seedminer + BannerBomb3 + unSAFE_MODE
+  + This method should only be used if you have a broken Internet Browser (Requires working shoulder buttons)
+1. [Fredtool](bannerbomb3): Seedminer + BannerBomb3 + Fredtool
   + This method is recommended if your shoulder buttons do not work
+1. [Pichaxx](homebrew-launcher-(pichaxx)): Seedminer + Pichaxx + HBL-Slottool
+  + This method is recommended if your DSiWare Management menu crashes and you have a broken browser
 1. [Pichaxx](homebrew-launcher-(pichaxx)): Seedminer + Pichaxx + Frogtool
-  + This method is recommended if your DSiWare Management menu crashes
\ No newline at end of file
+  + This method is recommended if your DSiWare Management menu crashes and you have broken shoulder buttons
```
lilyuwuu, line 57: This many options will inevitably confuse people

lilyuwuu (commented):
A few other comments, somewhat relevant to the one above

_pages/en_US/legacy-methods.txt
TheMasterOfMike: Never realized that before, so taking that into account, How exactly should it be listed/tested on either this page or on the Bannerbomb3 and unSAFE_MODE pages?

_pages/en_US/legacy-methods.txt
TheMasterOfMike: I partially thought that too but the problem I had was that I had no other feasible way I could easily do that. Linking directly to the Frogtool page would cause issues so I didn't try that. The only thing I could think of that could be better would be to make the entire PicHaxx thing one method and then lead to branching out methods on that page (which to be fair I basically already do in this PR) but then how would I exactly list it off on the legacy-methods page?

_pages/en_US/installing-boot9strap-(HBL-Slottool).txt
TheMasterOfMike: Responding to your original comment.

> I am against this change. People following the guide _normally_ will have very little need to ever go to this page, as browser will rarely end up broken (I've yet to see a case of a broken browser that is also on 11.13). While PicHaxx -> slotTool would technically be faster for people following video guides and only getting as far as pichaxx, it would be at the expense of confusing any non-video guide followers that are unfortunate enough to need to follow a legacy method. We also have a page off the guide if PicHaxx -> slotTool needs to be followed (that being said, it's my own site, so a bit of a conflict of interest here). It might be fine to create this page but not to link to it, so that we have the page on hacks.guide but people don't try to follow it as a path (because it really isn't a path that anyone on the main guide will ever need to follow).

We've already had people come in who were either in the middle of following the guide or had used an old guide, either for just userland, or they never finished installing CFW originally. Rather than linking them to frogtool (which is what the original hbl-usm does) it would be easier to link them to an HBL-Slottool page. Sure, your guide does exist but there are people who may not be able to use discord (either due to age restrictions in their country, country restrictions, parental restrictions, etc.) and they may not be able to easily find the guide, doing this not only benefits those people, but it resolves the need for your guide to be linked in the discord(though don't get me wrong your guide is still good regardless).

In regards to not having it be a visible path, I guess that is ok as long as frogtool isn't linked there either (cause why link frogtool there but not an easier method like HBL-Slottool), but at the end of the day, there is still quite a bit of people who have only userland access using an exploit such as PicHaxx or Steelhax, and leaving that on the guide as a path would prevent people needing to look or ask about what they should do if they already have an existing HBL exploit.

TheMasterOfMike:
Having to do this again because github decided to be github.

Responding to a significant part of your original comment.

> I am against this change. People following the guide _normally_ will have very little need to ever go to this page, as browser will rarely end up broken (I've yet to see a case of a broken browser that is also on 11.13). While PicHaxx -> slotTool would technically be faster for people following video guides and only getting as far as pichaxx, it would be at the expense of confusing any non-video guide followers that are unfortunate enough to need to follow a legacy method. We also have a page off the guide if PicHaxx -> slotTool needs to be followed (that being said, it's my own site, so a bit of a conflict of interest here). It might be fine to create this page but not to link to it, so that we have the page on hacks.guide but people don't try to follow it as a path (because it really isn't a path that anyone on the main guide will ever need to follow).

We've already had people come in who were either in the middle of following the guide or had used an old guide, either for just userland, or they never finished installing CFW originally. Rather than linking them to frogtool (which is what the original hbl-usm does) it would be easier to link them to an HBL-Slottool page. Sure, your guide does exist but there are people who may not be able to use discord (either due to age restrictions in their country, country restrictions, parental restrictions, etc.) and they may not be able to easily find the guide, doing this not only benefits those people, but it resolves the need for your guide to be linked in the discord(though don't get me wrong your guide is still good regardless).

In regards to not having it be a visible path, I guess that is ok as long as frogtool isn't linked there either (cause why link frogtool there but not an easier method like HBL-Slottool), but at the end of the day, there is still quite a bit of people who have only userland access using an exploit such as PicHaxx or Steelhax, and leaving that on the guide as a path would prevent people needing to look or ask about what they should do if they already have an existing HBL exploit.

_pages/en_US/installing-boot9strap-(HBL-Slottool).txt
lilyuwuu: The page can definitely *exist*. Not against that -- it's good to consolidate important information on hacks.guide. The thing is that putting it at the same level as frogtool will only *confuse* people.

The issue is that any main person following the guide (because remember, nobody will know which page to go to off the bat if they only have HBL access without asking us for help) will have no need to ever go Browserhax -> Legacy/Seedminer -> Pichaxx -> slottool.

_pages/en_US/legacy-methods.txt
lilyuwuu: It *can't* be tested

_pages/en_US/legacy-methods.txt
lilyuwuu: This only shows why it's bad to have this many confusing branches (can't do browserhax -> PICK 1/3: BB3-USM, BB3-dsidump, pichaxx; pichaxx -> PICK 1/2: pichaxx-usm, pichaxx-frogtool).

There comes a point where you have to sacrifice 'the fastest method' for what will make things more linear.

lilyuwuu:
The page can definitely *exist*. Not against that -- it's good to consolidate important information on hacks.guide. The thing is that putting it at the same level as frogtool will only *confuse* people.

The issue is that any main person following the guide (because remember, nobody will know which page to go to off the bat if they only have HBL access without asking us for help) will have no need to ever go Browserhax -> Legacy/Seedminer -> Pichaxx -> slotTool.

_pages/en_US/installing-boot9strap-(BH-Slottool).txt
TheMasterOfMike: So then why not move the section over to a page before? The only reason I even put it here is because eip originally put the frogtool page here, and I just went with it without (admittedly) using common sense to think "should this be placed here" and in retrospect it really should've been placed on the page before, both originally and now.

lilyuwuu: I'm not 100% against that, except that this guide is primarily stock to CFW and that the number of paths there are is already kinda high. The more links there are of potential scenarios (which, remember, are all edge cases), the more you might confuse the normal person (completely stock, wants to get CFW... or at least homebrew access).

TheMasterOfMike: I mean no offense but with this explicitly I believe, at least for the next couple of months, that it isn't a real edge case. We have already seen quite a few in NH Discord and it would make more sense to, at least for right now, list it on the page due to the somewhat high number of people with an existing HBL entrypoint. Would we remove it in the future? Possibly, but that would depend on how little it is being needed/linked to.

lilyuwuu: Remember that this page (referring to Browserhax 2020) is something that will be followed by **every person trying to mod their 3DS**. This text will not apply to the majority of people following this page (most people likely don't encounter issues, so they would never have any reason to join NH for support), making it an edge case.

NH is kind of a hub for issues, so it makes sense to have people with existing HBL entrypoints somehow end up there (Honestly, how it probably goes is people get HBL access, find hacks.guide somehow, get confused with "wait but I already have homebrew is that a kind of custom firmware", and get stuck and find the discord link).

That being said: I can see the point you're making, and since it's just an infobox it *should* be fine.

TheMasterOfMike: This should resolve this conversation then I believe.

_pages/en_US/installing-boot9strap-(HBL-Slottool).txt
TheMasterOfMike: Mind if I ask how exactly this could be worded better? I suppose the ordering of how everything is placed at the start could be altered but that probably wouldn't resolve the wording I am thinking.

_pages/en_US/legacy-methods.txt
TheMasterOfMike: So should a notice be added in both the pages of something along the lines of "if it freezes, then immediately returns to the home menu, you will be unable to continue with this method" or would that be too confusing to the end user? If it is, then how would we describe it in a way to how it would not confuse the end user whilst explain to the user that the method cannot be used due to a broken Nintendo DSiWare Data Management menu?

Oksana1983:
Thank you

_pages/en_US/homebrew-launcher-(pichaxx).txt
```diff
@@ -6,9 +6,9 @@ title: "Homebrew Launcher (PicHaxx)"
 
 ### Required Reading
 
-This method of using Seedminer for further exploitation uses your `movable.sed` file to gain access to the Homebrew Launcher using the PicHaxx exploit. This method requires you to already own (or download) the free "Pokemon Picross" game from the eShop.
+This method of using Seedminer for further exploitation uses your `movable.sed` file to gain access to the Homebrew Launcher using the PicHaxx exploit to allow us to preform further exploits. This method requires you to already own (or download) the free "Pokemon Picross" game from the eShop.
```
lilyuwuu, line 9: You misspelled 'perform'

_pages/en_US/installing-boot9strap-(HBL-Slottool).txt
```diff
@@ -10,11 +10,11 @@ In order to exploit the SAFE_MODE firmware of our system, we need to inject an e
 
 As we already have Homebrew access using an older entrypoint, we can use slotTool to do this.
 
-This specific method is intended to be used with older Homebrew Launcher access methods, such as PicHaxx and Steelhax.
+This specific method is intended to be used with only with older Homebrew Launcher access methods, such as PicHaxx and Steelhax.
```
lilyuwuu, line 13: "Used with only with"

_pages/en_US/installing-boot9strap-(HBL-Slottool).txt
lilyuwuu: it's just word order -- "If you don't currently have Homebrew Launcher access" is better

_pages/en_US/legacy-methods.txt
TheMasterOfMike: I partially resolved this (mainly the two PicHaxx things).

The main thing I still ask is how else should it be altered to specifically to fix the confusion issue?

_pages/en_US/legacy-methods.txt
lilyuwuu: It's kind of a 'we can't do anything' situation. We can't word it in a specific way very easily without creating too many options/confusing with the flash-black missing usm.bin, hence why the current wording is just a vague "if it doesn't flash red" with direction to come over to NH.

I think what you're trying to do with the guide is update the troubleshooting to be more up-to-date (to be more like the eiphax issues page). There is a time and place for in-line troubleshooting -- there comes a point where if we want to have all of the information on the guide we should link to a troubleshooting page/to the eiphax issues page.

TheMasterOfMike: I mean to be fair personally I find the troubleshooting the guide gives is kinda vague in many cases (see the whole issues with red bottom screen and no B9S install success stuff (which eiphax does not cover by the way) that we commonly get in NH, which is why I'm going to add another commit that would explain what to do in that situation when I gain access to a computer that I am able to use)

lilyuwuu:
There are still issues with this PR that make me hesitant to approve of it at this time, namely:
- It's gotten too messy / too broad. You're trying to accomplish too many things in one PR.
- Some of the changes you make are opinion-based -- we're going to need more opinions before we can continue.
- You tend to link to other pages a lot in the interest of the "ideal method", making the guide a lot more like a maze. While I'm not saying the guide doesn't *already* do this sometimes, you're making the issue *worse*.
- There are a great number of changes made by this PR, making it tough to review.
- You are still actively updating the PR, making it tough to review.
- You tend to over-describe things and word things weirdly as a result.

Here are my thoughts of the changes made in this PR as of `caf1426 Replace Legacy Methods mention with link straight to frogtool` accompanied by the changed text for my own and future reviewers' references. There might be repetition of points from earlier reviews. This comment is by no means a "be-all and end-all" of what you should and shouldn't have/do. **Additional feedback from other people is required for many of these points**.

# Homepage
1. CFW can be set up on any console on any version ~~(but may require additional tools / accessories for versions >11.3.0).~~
✔ Good change.

# Homebrew Launcher (Browserhax 2020)
2. ++ If you already have a current method of accessing the Homebrew Launcher, you should follow [HBL-Slottool](installing-boot9strap-(HBL-Slottool))
✘ Disagree with this change. 

Wording issues: `a current method` is confusing -- change to `already have access`. Also, if the person in question already has HBL access via Browserhax 2020, what then?

**Most people following this guide will be starting from a stock environment, as this guide has always been stock-to-CFW oriented.** If we want to account for people that already have HBL access, why aren't we also accounting for people that might have older CFW? That's the issue with trying to account for these cases -- you're opening a rabbit hole of what ifs.

A way around this might be to introduce a "do you already have homebrew access / custom firmware" on the get started page / do a Luma check / etc, but that's probably out of scope for this PR.

3. This should open a QR Scanner. If it does not, USM (the next part of getting custom firmware) will not work for you. You will need to use a legacy method, which will require [seedminer](seedminer)
**~** Reword this text, but otherwise a good change.

I get that you're trying to keep the text the same, but since all legacy methods are branched through Seedminer, this should probably be worded as `You will need to use [seedminer](seedminer)`.

4. Continue to [Installing boot9strap (BH-Slottool)](installing-boot9strap-(BH-Slottool))
{: .notice--primary}
✘ Needs review.

I don't personally agree with this `BH-Slottool` / `HBL-Slottool` naming scheme. Browserhax is still considered HBL access, and the only actual difference between the two pages is **one file** (`slotTool.xml`). 

I've already [shared my thoughts](https://github.com/hacks-guide/Guide_3DS/pull/1785#discussion_r483127528) on referring to the method as slotTool, and unless others say otherwise, we're keeping it as `???-USM`.

# Homebrew Launcher (PicHaxx)

5. This method of using Seedminer for further exploitation uses your `movable.sed` file to gain access to the Homebrew Launcher using the PicHaxx exploit to allow us to perform further exploits. This method requires you to already own (or download) the free "Pokemon Picross" game from the eShop.
✘ Disagree with this change

See point 7. 

6. If your device already has access to the Homebrew Launcher through PicHaxx, you can skip to ~~[DSiWare Dumper](dsidumper))~~ [HBL-Slottool](installing-boot9strap-(HBL-Slottool)).
✔ Agree with removing the link to DSiWare Dumper -- this was an oversight.
✘ Disagree with replacing it with HBL-Slottool (slotTool with XML)

See point 7.

7.  If you have working shoulder buttons, continue to [Installing boot9strap (HBL-Slottool)](installing-boot9strap-(HBL-Slottool))
If you do not have working shoulder buttons, continue to [Installing boot9strap (Frogtool)](installing-boot9strap-(frogtool))
✘✘ Strong disagree with this change

This has already been discussed in #hacking-general (and possibly earlier in this PR), but just so it's clear: 

**A very, very, VERY small minority of people following the guide will ever need to follow the sequence of `Browserhax 2020 -> Seedminer -> PicHaxx -> slotTool`.**

People will only end up following Seedminer if:
a) Browser is broken (incredibly unlikely -- cart updates to 11.13 do not exist)
b) Console is incompatible (KOR/TWN/CHN) with Browserhax (such a console would also not have access to Pokémon Picross)
c) **Shoulder buttons are broken**, rendering the person unable to follow slotTool.

Also see `The issue with the rest of the page`.

# Homebrew Launcher (Steelhax)

8. If your device already has access to the Homebrew Launcher through Steelhax, you can skip to [HBL-Slottool](installing-boot9strap-(HBL-Slottool)).
✔ Agree with this change. People will only ever end up on this page if they've been exploring or if they've been linked to it by an old video guide.

9. Removal of Steelhax instructions
✘ Needs review.

I know I was the one that suggested this, but other people need to agree with this change. My points are that there is effectively no need for steelhax instructions these days, and that the page should be updated to reflect that steelhax is dead (while keeping the page live so that there aren't dead links from old video guides or something).

10. ++ This is an outdated method and is no longer working or supported. Unless you have the files to skip to HBL-Slottool, you should start from [BrowserHax 2020](homebrew-launcher-(browserhax-2020)).
**~** Needs rewording.

Possible wording issues: `outdated method and is no longer working or supported` seems a bit odd to me, but I might be wrong.

Wording issues: The text above (point 8) says that people with steelhax should skip to HBL-Slottool, but this point says that unless they have 'the files to skip to HBL-Slottool' (so presumably steelhax files?) that they should be following Browserhax 2020. This can be a bit hard to understand.

I'd reword this to something like:

`This method is no longer supported as it has become outdated. If you don't already have access to the Homebrew Launcher via Steel Diver: Sub Wars, you should start from [BrowserHax 2020](homebrew-launcher-(browserhax-2020)).`

# Installing boot9strap (BH-Slottool)
I disagree with the name of the page -- see point 4.

11. As we already have Homebrew access using BrowserHax 2020, we can use slotTool to do this.
**~** Nothing necessarily *wrong* with this change, just seems unnecessary. I'd also use `through` instead of `using`.

12. ~~If you have Homebrew Launcher access through another entry point, you should use [Frogtool](installing-boot9strap-(frogtool)).~~
**~** It is somewhat contradictory to point 2 to disagree with this change, but I sort of disagree with it as it provides a potentially useful failsafe. I haven't seen anyone get paticularly frazzled by this text, anyhow.

13. If your (Right/Left Shoulder), (D-Pad Up) or (A) buttons do not work, you will need to use a legacy method, which will require [seedminer](seedminer).
**~** Reword this text, but otherwise a good change. See point 3. 

14. If it just freezes at a red bottom screen, verify that usm.bin is on the root of the SD Card, then retry from the start of this section.
**~** Reword this text -- it isn't guide-style ("just freezes"). Otherwise a good change.

`If you get stuck on a red screen, verify that usm.bin is on the root of the SD card, then retry this section`

(These bullet points tend not to end in a period)

# New page (Installing boot9strap (HBL-Slottool))
I disagree with the name of the page -- see point 4.

15. As we already have Homebrew access using an older entrypoint, we can use slotTool to do this.
✘ You don't need to specify that the homebrew access entrypoint is older.

16. If you don't currently have Homebrew Launcher access, you should use [BrowserHax 2020](homebrew-launcher-(browserhax-2020)).
If you currently have Homebrew Launcher access using BrowserHax 2020, you should follow [BH-Slottool](installing-boot9strap-(BH-Slottool)).
✘ Disagree with this text.

Consistency issues: `you should use` and `you should follow`.

People shouldn't end up at this page if they're able to follow Browserhax 2020/slotTool without XML. Putting this text here will be unnecessarily confusing.

17. These instructions work on USA, Europe, Japan, and Korea region consoles as indicated by the letters U, E, J, or K after the system version.
**~** People in Korea wouldn't have a free HBL entrypoint, so unless BannerBomb3 can get HBL working, it's kind of pointless to mention it (or even mention regions, for that matter).

18. Copy the contents of the `slotTool` folder from the unSAFE_MODE `.zip` into the `3ds` folder on your SD Card
✘ Reword this text.

You can just copy the `slotTool` folder into the `3ds` folder; you don't need to specify the contents.

19. If it just freezes at a red bottom screen, verify that usm.bin is on the root of the SD Card, then retry from the start of this section.
**~** Reword this text -- see point 14.

# Installing boot9strap (usm)`

20. ~~If your (Right/Left Shoulder), (D-Pad Up) or (A) buttons do not work, you will need to use a [Legacy Method](legacy-methods). For assistance with this matter, join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp) and ask, in English, for help.~~
✘ There isn't really a need to remove this text.

# Legacy Methods

21. ++ If the system is found to be unminable, and formatting the console does not change that, you will either need to follow [BrowserHax 2020](homebrew-launcher-(browserhax-2020)) or follow an All Versions method.
**~** Reword this text, otherwise neutral on this change.

`If the system is found to be unminable, and formatting the console does not change that` is a bit odd to read.

People shouldn't get to this page unless they're unable to follow Browserhax 2020, but it's not the end of the world if it's mentioned here.

# The issue with the rest of the page
There's a fundamental issue with guide structure that this PR exacerbates: The structure of the guide for 11.13 has become very maze-like.

This is the guide structure as it stands (Blank boxes are part of the main branch, red/strikethrough boxes are no longer supported):
image
We can already see that it's getting a bit confusing, with multiple branches for different/"ideal" circumstances.
This is the guide structure you are proposing with this PR:
image
This fragments the guide into an even bigger maze.
Here's the structure I'd like to propose (Yellow boxes are still supported, but not linked on the main branch):
image
My reasoning is this:
- Installing boot9strap (usm) now has just one use case: the console is unable to use browserhax (either because the browser is broken, or because the console is Korean). While this seems like an important point, the guide does not currently have a Korean translation, and Korean folks tend to follow [a guide on naver](https://blog.naver.com/PostView.nhn?blogId=dongs0530&logNo=221988630243) instead. Under this proposition, the page would still exist, so if anyone ever needed help with it they'd be directed to the appropriate page from NH.
- BannerBomb3 (DSiDump) now has even smaller use cases: either the console is Korean AND has broken shoulder buttons, or the console is unable to access the eShop to download Pokemon Picross for whatever reason. While it'd technically be *quicker* for people with working DSiWare Management menus, I think it's fine to sacrifice time saved in favour of simplicity.
- See point 7 regarding PicHaxx+slotTool.

Therefore, for most people, this alternate guide structure for broken shoulder buttons will suffice.

# Region Changing

22. ++ After this process, it is extremely likely that you lose the ability to access the eShop on your new region. If you do retain eShop access, changing the region back to the original region for any reason will break the eShop for the original region.
**~** Reword this text, but otherwise a good change.

It is extremely likely that you **will** lose the ability to...

`If you do retain eShop access, changing the region back to the original region for any reason will break the eShop for the original region.` is a bit janky -- it sort of implies that not changing the region will somehow keep the eShop working on the original region.

# Seedminer

23. To install boot9strap on your device using a legacy method, we exploit an oversight in the security restrictions applied to "DSiWare" applications.
✘ I don't see why you need to specify that this is a legacy method.

24. Removal of DSiWare Management Test
✔ Agree with this change -- it was based on misinformation.

25. The Legacy Methods subheader
**~** Besides `These methods are outdated and not recommended unless the reason listed applies to you.` being an unnecessary/oddly worded change, this is linked with `The issue with the rest of the page`.

There's also no reason to have Seedminer link to legacy methods anymore. The methods (no matter how many we decide on having) should just be moved to the one page.

# Site Navigation
✔ No issues with adding pages to navigation.

TheMasterOfMike:
Let me go through this and explain a couple of my own thoughts with this.

## Section I - Issues with the PR itself

> You're trying to accomplish too many things in one PR. / You tend to over-describe things and word things weirdly as a result.

I can admit that's clearly visible and I realize this is an issue I have, it's an issue with how I do stuff in general, this isn't the first time I've been told I attempt to accomplish too much or over-describe stuff and come off weirdly due to that.

> You tend to link to other pages a lot in the interest of the "ideal method", making the guide more like a maze.

I'll go through this a bit more when I get into the issues with your "The issue with the rest of the page" section. But basically, I can _to an extent_ see what you mean by that. My original goal, as I have stated was to make the guide consistent, accurate, and not link to outdated method just to make guide making simpler. While I personally feel I succeeded with the majority of my intentions with this PR in general, there are definitely issues with the way I did things (and I will address this later)

The rest of these things, while I could explain them, I do not personally believe it would be beneficial to do so.

## Section II - Issues with the BH2020 page

I'm not going to always quote these because some either would be an extremely lengthy quote or it wouldn't be beneficial to do so

Regarding your first point of this section, I believe the main issue is more of a wording issue than an actual issue with it being linked there itself. Personally, I believe that adding it to the get started page would only complicate matters more, and only linking it from the steelhax page (which I will get into later) seems like a somewhat bad idea.

Regarding your second point of this section, I believe that saying "You will need to use a legacy method, which will require seedminer." makes more sense and actually goes through the intent of doing seedminer (also  then saying "You will need to use seedminer."

## Section III - The Naming Convention

I decided to put this as a separate section only because it's a significant point of this entire thing.

The main reason I decided to rename the pages to both BH-Slottool and HBL-Slottool is because (this is personal opinion) they are, in my eyes, the best names that would be consistent across each of them.

If we didn't use BH-Slottool and HBL-Slottool as the names, what would we name the HBL-Slottool? HBL-XML-USM? That sounds like a significantly worse name and would just sound confusing/wrong, sure the file difference is just one file but at the end of the day if we just made them the same page as well, we would just end up with a lot more confusion.

## Section IV - Issues with the PicHaxx page

(This is actually only two points since you pretty much only have two points as your reasoning)

> **A very, very, VERY small minority of people follow the guide will ever need to follow** (sequence)

> People will only end up following (that sequence) if (three things listed)

In retrospect, this was something that I should have very clearly noticed and thought of, and it's quite an obvious thing I should have noticed originally.

(Not actually going to cover the steelhax page by the way)

## Section V - Issues with BH-Slottool

(Will fix the wording issues regarding all but the second point)

Regarding the second point, (removing the mention of Frogtool) I moved a similar equivalent to HBL-Slottool (which I'll address later. And secondly, while it hasn't necessarily affected helpee's, helpers (not staff helpers but you know what I mean) meanwhile have almost (and sometimes) have linked to Frogtool (with the users having working buttons), even though we can just link them to Slottool (especially since they are in NH discord)

## Section VI - Issues with HBL-Slottool

Regarding the first point, the entry point is technically older and the point you're making almost sounds like somewhat of a nitpick, but eh moving on.

(Will fix the consistency issues with the rest of the points)

## Section VII - Issues with the bb3+USM page

Regarding the only point, why leave the mention there if they have to go through legacy methods to get to the page anyways, which would make them aware of said requirement.

## Section VIII - The maze-like structure

> There's a fundamental issue with the guide structure that this PR exacerbates: The structure of the guide for 11.13 has become very maze-like.

I hope you don't mind me asking why it took this PR for this to finally be very noticeable? This has been very visible in my eyes ever since BH2020 was added to the guide. I didn't help the issue, but it almost seems like this issue wasn't noticed until my PR was made despite it actually having been an issue.

Regarding your ending chart, I have a couple of things I want to point out here.

1. I personally have been in favor of removing bb3+USM from the legacy methods page but since it was there when it was originally put in I never bothered with fixing that (and also didn't think of the actual lack of need for said edge cases).

2. (Time for personal opinion here) I'm in personal favor of linking to bb3+Fredtool if they don't have working buttons.

Let me explain here.

- We don't have to worry about people being unable to use torrent clients or not understanding what they even are.
- We don't have to worry about needing an NNID to do this (though to be fair you could just do BH2020+Frogtool but that would complicate things)
- It's unlikely people with broken buttons have uninstalled CFW or managed to break their TWLFirm, which would make inability to do Fredtool an edge case.
- If the injector was an issue, we could always just directly link to the Alternate Injector (you have to do this for PicHaxx anyways)

Sure, there are edge cases where bb3+Fredtool may not be the best option, but again, they are **edge cases**.

(Side note, what should we do about the DSi Dumper page if we're marking that in red/no longer supported)

(Going to fix the issues with the region change page as well... moving on)

## Section IX - Issues with the Seedminer page

> I don't see why you need to specify that this is a legacy method.

Because it is required only if you're doing a legacy method.

Also, we could just link seedminer to either bannerbomb3 -> Fredtool or PicHaxx -> Frogtool (that's still a decision that needs to be determined)

## Other notes

My honest opinion with this entire guide, based on the things that you brought up as well as from what I have noticed, is that, (with all respect to the people who work on the guide) there are a multitude of flaws, a good chunk of which are flaws that could have been easily resolved in the PR for BH2020 or even prior to that. (Such as the Steelhax page still existing like it was not outdated, the PicHaxx guide still mentioning DSiWare Dumper, or the home page still mentioning additional tools would be required for >11.3.0)

lilyuwuu:
> Issues-PR: My original goal, as I have stated was to make the guide consistent, accurate, and not link to outdated method just to make guide making simpler.

OK, yeah, but there are logical times and places for these things. Just because USM-based methods are the premier methods doesn't mean that they should logically always be used. Frogtool and Fredtool-based methods are functionally identical in terms of the actual "dangerous part" of installing CFW. Of course the ideal would be to have people be linked to whatever the ideal method actually is, but there is just no realistic way to word this without making the guide unnecessarily confusing.

> BH2020: Personally, I believe that adding it to the get started page would only complicate matters more

Yeah, that's why we haven't and aren't doing it. 

> BH2020: only linking it from the steelhax page (which I will get into later) seems like a somewhat bad idea.

You didn't get into this later.

> BH2020: Regarding your second point of this section, I believe that saying "You will need to use a legacy method, which will require seedminer." makes more sense and actually goes through the intent of doing seedminer (also then saying "You will need to use seedminer."

I'm not 100% sure what you mean here. It almost feels like we're *shaming* people for using legacy methods.

My point is that we don't need to have a Legacy Methods page anymore. What's the point of having the Seedminer page (already not linked as a main guide path) only link to "hey maybe you shouldn't use legacy methods"?

> Naming Convention: The main reason I decided to rename the pages to both BH-Slottool and HBL-Slottool is because (this is personal opinion) they are, in my eyes, the best names that would be consistent across each of them.

Yeah, and I personally feel the name feels a bit janky. It's difficult to name these things. As much as I'd like to call this a trivial matter, these are pages that can be potentially linked elsewhere, so renaming things a bunch isn't ideal.

> BB3-USM: Regarding the only point, why leave the mention there if they have to go through legacy methods to get to the page anyways, which would make them aware of said requirement.

Fair point, overlooked this.

> Maze: I hope you don't mind me asking why it took this PR for this to finally be very noticeable?

That's because before the browserhax PR, the structure was a lot simpler:
image

The browserhax PR's proofreading was done in about a day by two people -- I didn't notice that USM had just been shoehorned into the Legacy Methods page.

Issues only get noticed **if they're brought to our attention**. So you're right, we hadn't noticed this. But if you noticed it, why didn't you make an issue about it?

> Maze: I'm in personal favor of linking to bb3+Fredtool if they don't have working buttons.

The points you make are fair. While I'd argue that torrent clients are pretty easily installable (people are already pushed to install 7-zip, after all -- it was *essential* for the longest time until Luma finally switched to ZIP releases... School Chromebook™ users would encounter issues in both circumstances) and that most people should have NNIDs, it's also not the end of the world to link both BB3 and Pichaxx. The only annoying thing is that in the past, people would always ask which of BB3 or Pichaxx was "better", but since this is all within the Legacy section, it should be fine. Could be distinguished simply as if your DSiWare Management menu works or doesn't work... except that this is difficult to outright test.

> Maze: (Side note, what should we do about the DSi Dumper page if we're marking that in red/no longer supported)

Keep it in relative obscurity, I guess? Nobody ever looks at the page and I can't see many scenarios where a video guide would link to it.

> Seedminer: Because it is required only if you're doing a legacy method.
Also, we could just link seedminer to either bannerbomb3 -> Fredtool or PicHaxx -> Frogtool (that's still a decision that needs to be determined)

That's probably best.

> Other: there are a multitude of flaws, a good chunk of which are flaws that could have been easily resolved in the PR for BH2020 or even prior to that.

We can't resolve flaws that we're not aware of. Because Legacy Methods won't get much traffic (and most people that need to follow it that end up at NH will just get linked to the proper respective guide), it's difficult to find these flaws, and it's also difficult to gauge how many people get confused (and potentially turned away from the guide) because of them.

GlaZedBelmont:
Talked with the guide team about it, while some points are good, many of them are unnecessary, feel free to make a PR with the good points (see Lily's last comment).

PR #1793, opened 2020-09-23
fix DO_NOT_LAUNCH.firm link in italian version of finalizing setup

radostin04:
yes this isnt crowdin but bleh. the issue is caused by the link pointing to /assets/do_not_launch.firm. on the english version, this points to 3ds.hacks.guide/assets/do_not_launch.firm which is valid. on translations, this point to 3ds.hacks.guide/[language]/assets/do_not_launch.firm which is not valid. i imagine there's a more proper fix for this but i reckon this will work for now


lilyuwuu:
This issue is likely present in all languages. Language changes can't be PRed this way because they'd get overwritten by crowdin, iirc.

#1789 should resolve this by ditching the file altogether

lilyuwuu:
Closing; DO_NOT_LAUNCH.firm has been removed from the guide, should be fully removed once the new crowdin updates are merged

PR #1795, opened 2020-09-25
Update homebrew-launcher-(pichaxx).txt

adrot:
Removed extra parentheses.


lilyuwuu:
There's nothing specifically wrong with this change, it's just that we don't actually use dsidumper anymore (this text was overlooked in the past)

lilyuwuu:
Closing this -- rest of the text needs to be updated anyway. Planning to do this when reorganizing the Seedminer branches of the guide to be less of a maze.

PR #1797, opened 2020-10-09
Get Started table revamp

lilyuwuu:
This PR will primarily change the _Get Started_ table to make it easier for users to understand. While #1782 accomplishes this by replacing the table with a dropdown for people with javascript enabled, a dropdown may not be the ideal solution / could overwhelm people due to the sheer number of 3DS firmware versions. Still need more opinions on it, though -- this PR could be combined with the other one to make this the table should the dropdown fail.
image
Desktop view (Current on left, PR on right)
[Mobile view](https://i.imgur.com/00AtGrB.png)

Changes in this PR:
- Remove Cart Update text, as it is no longer relevant
- Column width is slightly resized due to weird alignment issues. Slight column alignment issues still seem to appear when Chrome tries to simulate the screen of a mobile device, but don't seem to appear on an actual mobile device.
- The order of the versions is reversed, as people are most likely to be on 11.13 when they are modding their consoles, so it makes sense for it to be the first thing listed.
- Add "latest" text to 11.13 to make it clear that it is the latest firmware version.
- Change text for 1.0-2.0 and 11.4-11.12 to explicitly tell users to update to the latest version. _All Versions_ methods are rarely the ideal method for anyone, and the text primarily confuses people on 11.4-11.12 under the impression that their firmware is easier to mod.
- Separate "All Versions" into a second table. Partially a stylistic choice, partially in an effort to physically separate those two methods from the guides that most will be following. Still not 100% sure of this change though, might be more trouble than it's worth or end up causing more confusion.
- Change the style of the two methods in "All Versions". At present, the "requires a compatible flashcart" / "requires soldering" text is smaller and potentially harder to read.

EDIT 2020-11-18: Added latest text to guide in earlier commit. Changed text to apply to 11.14.


lilyuwuu:
Merging this PR with #1812.

PR #1798, opened 2020-10-13
Uninstall CFW: Improve safety test

lilyuwuu:
The following PR will make the Uninstall CFW guide a bit safer and hopefully try to prevent broken DS mode and/or System Settings, which normally render it harder/impossible to reinstall CFW on the console.

The actual important changes of this PR:
- Switch safety test fail to a red warning box rather than a bullet point
- **Add Functionality Test section** to test DS download play and system settings/keyboard

Slight wording changes:
- `boot9strap` capitalization changed to be consistent
- ""... after this guide" -> "... after following this guide"

Potential issues:
- **This PR does not currently change the fact that hidden DSiWare that doesn't appear in Data Management is left installed.** I'm not 100% sure of a good way to check for that, but it's a bit less important than DS mode and System Settings/the keyboard


radostin04:
Wouldn't going into DS mode while in stock cause you to reboot into CFW after exiting?

lilyuwuu:
Hm, you're right, didn't think about that. :/
maybe add instructions to set native firm as boot.firm (maybe create a script that would set native firm as your boot.firm?... Potentially dangerous though) -- Still thinking that testing these things is good but want to do so in a way that won't make things worse somehow

techmuse8:
Something I thought was this that would be reasonable and not as confusing to the end user at the same time, couldn't you make a script that could automatically extract the native firm (would have to use a different script if you had a new or old model), rename the .firm to native.firm (assuming gm9 scripts could do that) put it in sdmc:/luma/payloads, and when they exit out of DS mode to have them hold start (where I'm sorta against this) to go to the chain loader and boot back into native firm?

476MHz:
One could also pack it all into one script, and have it detect new or old by looking at the folder names (specifically the folder directly before content), and moving the right file accordingly.

techmuse8:
Could use new or old detection by using this https://github.com/d0k3/GodMode9/blob/master/resources/sample/HelloScript.gm9#L38 then if its new have it look at "SYSNAND CTRNAND > title > 00040138 > 20000002 for New and vice verse for Old.

ItsPizzaTime1501:
NATIVE_FIRM as boot.firm wouldn't be too problematic as far as i know, the only thing that i would add is a warning telling the user not to update while in stock because with boot9strap as firm0 updating would call [AMPXI:InstallFirm](https://www.3dbrew.org/wiki/AMPXI:InstallFIRM). I'm pretty sure you don't get update nagged when you have boot9strap as firm0 while in stock although it's better to be safe than sorry.

techmuse8:
Besides in most cases wouldn't the user already be on the latest version? (Unless they had missing system titles then update nags)

lilyuwuu:
This *should* be functional now? The script looks like this:
image

Existing issues, potentially out of scope for this PR:
- Hidden DSiWare is still not uninstalled. I'd rather people avoid going into FBI > Titles because they could further mess things up there...
- DS Connection Settings is not tested. Part of this is because I'm not sure of reboots and part of this is because DS Connection Settings is overall a less useful feature than DS Download Play.
- The safety test script is entirely in English, so people following the guide in a different language would have trouble actually *reading* the script.

techmuse8:
For the 2nd issue I can confirm that it does reboot to CFW when in stock mode.

MechanicalDragon0687:
> * Hidden DSiWare is still not uninstalled. I'd rather people avoid going into FBI > Titles because they could further mess things up there...

I made an app for that, but I agree that is partially out of scope. We could recommend running the app and system formatting to remove any invalid dsiware, homebrew apps, and invalid tickets (Which would prevent updating games that have invalid update title tickets, for instance, if faketik was run before opening eshop or without installing eshop tickets via fbi)

Actually the ticket issue could be a big issue for uninstalled systems. If there are fake tickets for any legitimate software or updates, there would be no way to resolve this, after uninstall, without a system format.  There does not seem to be a way to tell which tickets are legitimate at this time in order to manually remove them. 

It is possible to remove all unused tickets, but this has people go into FBI->tickets (though i havent had issues in channel with directing people to do this), but it would still require removing all updates/dlc then redownloading them.

> * DS Connection Settings is not tested. Part of this is because I'm not sure of reboots and part of this is because DS Connection Settings is overall a less useful feature than DS Download Play.

We talked about this, and this can be tested in cfw mode without issue and would make sense to check.

_pages/en_US/uninstall-cfw.txt
```diff
@@ -6,65 +6,128 @@ title: "Uninstall CFW"
...
-  + If you do NOT boot into the regular 3DS Home Menu, continuing with these instructions **WILL** brick your device. You should join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp) and ask (in English) for someone there to assist you
-  
-#### Section II - Prep Work
+  + Note that, if you had to fix the RTC date and time, you will have to fix the time in the System Settings as well after following this guide
+1. Press (Home) to bring up the action menu
+1. Select "Scripts..."
+1. Select "safety_test"
+1. Read the text on-screen and press (A) to continue
+1. You should boot into the regular 3DS Home Menu. If you do, continue these instructions
+  + If you reach an error screen, uninstalling CFW **WILL BRICK YOUR DEVICE!**
```
radostin04, line 72: This should probably mention the possibility of a black screen as well, possibly "If you reach an error screen, or your device does not turn on..."?

_pages/en_US/uninstall-cfw.txt
```diff
@@ -6,65 +6,128 @@ title: "Uninstall CFW"
...
+1. Follow the prompts and allow the program to uninstall 
+1. Once the process has succeeded, exit the Homebrew Launcher and power off your device
+
+#### Section VI - System Format
+This section will ensure that all illegitimate tickets are removed, allowing eShop to work normally. This will remove all content from the 3DS and log you out of your NNID. Keep in mind that your console's encryption key will be shuffled, meaning that any old data will be rendered inaccessible, even if you have a backup of your SD contents.
 
-#### Section III - Running Uninstall Script
+1. Power on your device
+1. Open the System Settings application
+1. Navigate to Other Settings -> Next Page (four times) -> Format System Memory
```
radostin04, line 125: I think that on certain 3DS models there's only 4 pages of "Other settings", and not 5? I might be wrong on this though

lilyuwuu: JPN/USA N3DS XL have five pages, USA O3DS XL has four. Will amend

PR #1817, opened 2020-12-09
installing boot9strap (soundhax) -> 1.0.0 to 11.3.0

lilyuwuu:
**Description**
Use `Installing boot9strap (Soundhax)` for all versions between 1.0.0 and 11.3.0, using universal-otherapp.

Changes in this PR:

get-started:
- Remove `This means that, for example, the "from 9.0.0 to 9.2.0" row includes 9.0.0, 9.1.0, and 9.2.0.` because it doesn't fit the current guide structure
- `Installing boot9strap (Soundhax)` -> from 1.0.0 to 11.3.0

installing-boot9strap-(soundhax):
- Remove USA/EUR/JPN/KOR region text ("all regions")
- Remove cart update notice
- Adjust instructions accordingly for use with universal-otherapp

site-navigation:
- Switch popular page from HBL-Soundhax to B9S-Soundhax

### DIFF
--- a/_pages/en_US/get-started.txt
+++ b/_pages/en_US/get-started.txt
@@ -6,7 +6,7 @@ title: "Get Started"
 
 ### Required Reading
 
-Select the appropriate page for your version from the chart below. Note that the "from" and "to" fields are inclusive. This means that, for example, the "from 9.0.0 to 9.2.0" row includes 9.0.0, 9.1.0, and 9.2.0.
+Select the appropriate page for your version from the chart below. Note that the "from" and "to" fields are inclusive.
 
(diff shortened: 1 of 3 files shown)


NightYoshi370:
Shouldn't we delete `Homebrew Launcher (Soundhax)`?

radostin04:
I argue that there might be cases where we'd want to keep that page, although it should be removed from everything but site navigation.

lilyuwuu:
> I argue that there might be cases where we'd want to keep that page, although it should be removed from everything but site navigation.
My thoughts exactly

PR #1821, opened 2020-12-19
Implement ssloth browser check bypass, new-browserhax-xl and safecerthax

lilyuwuu:
This PR implements new-browserhax-xl (n3ds) and safecerthax (o3ds). **There are drastic structural changes!**

Current guide structure:
image

Modified guide structure:
O3DS -- 
image
N3DS -- 
image

**What this PR changes**
- Modify existing `Installing boot9strap (Browser)` for use with new-browserhax-xl + universal-otherapp
- Create `Installing boot9strap (safecerthax)`
- Create `Installing boot9strap (SSLoth-Browser)`
- Split Get Started into three pages, like the guide pre-2018:
    - Get Started -- select New 2/3DS or Old 2/3DS
    - Get Started (New 3DS)
    - Get Started (Old 3DS)
- Update Site Navigation to accommodate these changes

**Things currently missing in this PR**
- ~~Proper step tracking to accommodate for the new pages~~ Tracking has been fixed
- ~~Description of exploit in required reading sections -- potentially unnecessary~~ A short blurb has been added to show what consoles/versions the exploits are compatible with, but no technical information.


NightYoshi370:
Since the main guide structure no longer accesses The Homebrew Launcher, could we just rename the page title to contain just the method name? Installing Boot9Strap is already known via the Home Page, so the only new thing that people would know is how.

NightYoshi370:
Actually, scrap that. I can see why you would want to keep it (other pages in the guide exist). I guess my main point about giving the exploit title more attention still stands, so why not try something like this:

- "safecrthax -- Installing B9S"
- "Browser -- Installing B9S"
- "SoundHax -- Installing B9S"

lilyuwuu:
Using `method -- way` would likely create the monstrous URL of `https://3ds.hacks.guide/method----way`, which is IMO worse than our current issues with brackets. :P 

I'd rather not break from guide convention unless we set up redirects. We're certainly free to set up a *new* guide convention, but that would mean that we would change basically all of the pages for the sake of consistency.

NightYoshi370:
You could manually set a url via the `permalink` attribute.

GlaZedBelmont (approved):
I see no issues with this PR, good job!

MrNbaYoh:
Hey, I updated my servers to bypass the browser update check that prevents users from exploiting browserhax on version < 11.14, todo so you only need to use the safecerthax DNS
So maybe you could add a section (SSLoth + browserhax) for n3ds users on version 11.4-11.13?

lilyuwuu:
Added `Installing boot9strap (SSLoth-Browser)` page -- please re-review to ensure there aren't any errors or missing/unnecessary steps. :)

techmuse8:
In SSLoth-Browser, since the lowest it supports is 11.4, shouldn’t there be a notice about cartridge updates “removing” the browser?

eip618 (approved):
Looks good to me. Now we all just have to learn these ourselves....

PR #1825, opened 2020-12-26
Retool seedminer based on new releases

lilyuwuu:
Changes in this PR:
- Create `Installing boot9strap (PicHaxx)` -- PicHaxx+universal-otherapp.
- New Seedminer guide structure is to pick either `Installing boot9strap (PicHaxx)` (recommended) or `BannerBomb3 + Fredtool`.
    - PicHaxx+Frogtool no longer has any use, as universal-otherapp should be usable in all cases where HBL+Frogtool is.
    - Installing boot9strap (usm) no longer has any use, as the only use case for Seedminer is broken shoulder buttons on O3DS.
- Add "Seedminer has been replaced by easier/safer methods" warning to every Seedminer page.
- Remove `Legacy Methods` page.


TheMasterOfMike:
Decided to respond once here, I probably won't read any responses to what ends up being stated, and I probably won't respond any further unless I see reason to, as I've pretty much changed my focus to other things... that being said, I want to point some things out:

- There really is not any reason to suggest BannerBomb3+Fredtool anymore as a legacy method, apart from edge cases. This is due to the fact that, unless you somehow cannot access PicHaxx, somehow cannot access the eShop, or somehow cannot access the internet due to a hardware issue, there really is not any reason for this method to remain on the main structure (though keeping it on site navigation makes sense)
- SSLoth works on old systems as well, and as well, old-browserhax was recently updated with universal-otherapp as well, this makes it so that we don't even have to recommend seedminer for pre-11.14 systems with broken shoulder buttons (this may seem like an edge case, but I would not be surprised if many old model systems (which are probably older systems with more wear/issues as well) are probably still on something like 11.13.0-45, considering that that was the version out for almost an entire year. Due to that, it probably would make more sense to attempt to avoid miner-methods, as, if the console is un-mineable, they're screwed if they update unless they either repair the shoulder buttons/d-pad or they get an NTRBoot cart, which both could be costly depending on the area they are located in)

lilyuwuu:
> There really is not any reason to suggest BannerBomb3+Fredtool anymore as a legacy method, apart from edge cases. 
Seedminer *itself* is an edge case now. For edge cases, people will likely end up coming to NH, but it's better to have the pages linked.
The edge case I'm thinking of is no access to eShop, which actually just happened earlier today.

If people have a wifi module issue they need one-on-one help to mii mine or something.

> SSLoth works on old systems as well, and as well, old-browserhax was recently updated with universal-otherapp as well, this makes it so that we don't even have to recommend seedminer for pre-11.14 systems with broken shoulder buttons (this may seem like an edge case, but I would not be surprised if many old model systems (which are probably older systems with more wear/issues as well) are probably still on something like 11.13.0-45

Already considering re-wording the edge cases, maybe won't mention Seedminer specifically in the safecerthax page. The SSLoth page already works on both N3DS and O3DS.

TheMasterOfMike:
Probably is the best call to just make it so that if you cannot do the main method, go to (NH) to be given options/assistance, especially given that, at this point, the only reason to use Seedminer is basically just a bundle of edge cases.

That being said, I want to point something out:

If we assume, let's say, we don't go through with that idea you and I both now bring up, and we keep seedminer/legacy methods mentioned on the safecerthax page. In that circumstance, it would still make sense to only mention Installing boot9strap (PicHaxx), and here is why:

- Could keep the legacy method page and make it an almost chart of sorts, mentioning both SSLoth and Seedminer, without making Seedminer result in a further branch (and further complicating things).
- Would allow either making Seedminer directly lead into PicHaxx or merge them (latter is a stretch, but thought I'd bring up the possible idea)
- Sure we don't recommend it, but at the same time, if one say has previously done the method before, it would be fair for one to assume it would be "easier" for them to do that than to do the better/easier method(s) and therefore, not do the one that is blatantly recommended to them (though admittedly you could argue about user choice about a method)
- (Also doesn't require an illegal version of Flipnote Studio and doesn't rely on B9STool (which is generally worse than the one PicHaxx now uses))
- While it is possible the idea of only mentioning one could lead to oversimplifying (which is fair) I also would like to mention that the tab that tells you that if stuff doesn't work to use Seedminer already keeps note of that you should go to NH for assistance, (so accounting for even less likely edge cases would probably be redundant):

> If your (Right/Left Shoulder), (D-Pad Up) or (A) buttons do not work, you will need to use an alternate method, like Seedminer. **For assistance with this matter, join Nintendo Homebrew on Discord and ask, in English, for help.**

lilyuwuu:
Yeah, as much as I'd like to be able to avoid sending people to NH whenever possible it's just not feasible to integrate all of this information into something people would be able to easily digest.

TheMasterOfMike:
So bit of a thing... you may not actually be able to complete the SafeB9SInstaller part with a dead d-pad (and probably A button) meaning Seedminer and legacy methods may need to return to being mentioned on the new 3ds side of things (or better yet just use the NH mention mentioned earlier in this PR thread).

Context: New Nintendo 3ds XL, 11.13.0-45U, due to a fucked up repair, the d-pad is completely dead, which previously also locked me out of all the main methods ever since UnSafeMode+bb3 was implemented of CFW'ing this console (also have been lazy to be honest to just do Fredtool).

I went ahead and followed SSLoth+new-browserhax, (this also applies to newbrowserhax-xl as well, but not relevant to this case directly) then I got to the SafeB9SInstaller part... prompted me with the key combo part, attempted using both the circle pad and the c-stick, no response came on SafeB9SInstaller, which makes it seem like it only reads direction-based inputs from the d-pad... which I cannot use due to the reasons above.

Point is... Seedminer is actually still necessary for a decent handful of new model consoles, not to mention... this also means I was wrong in downplaying the need of bb3+Fredtool... and also means we may have (admittedly huge edge case) a small need for PicHaxx+Frogtool...

This also may mean Seedminer is no longer a complete full series of small edge cases, but rather just slightly less of a need compared to when BH2020 and old-browserhax/new-browserhax were the main methods.

lilyuwuu:
A dead d-pad is a large edge case?

???

TheMasterOfMike:
I'm confused what you're saying. are you saying it's not a huge edge case or is a huge edge case?

lilyuwuu:
It's not a huge edge case at all. I cannot recall off the top of my head someone with a broken d-pad. Maybe it'll be a more prominent "issue" with SafeB9SInstaller gaining more usage, but in the past 24 hours there hasn't been a single case so it's definitely not that widespread. A couple of other thoughts:

- **GodMode9** also forces you to use d-pad for file unlocking by default.
- GodMode9 has an alternate build where you can hold a button. I'm guessing the same thing is possible for SafeB9SInstaller if it becomes a large enough edge case (which I highly doubt). Or we could probably even chainload into a (potentially older) build of GodMode9.

TheMasterOfMike:
Eh fair, the only major reason I bring it up is this part of the mention on the old 3ds side of things.

> If your (Right/Left Shoulder), **(D-Pad Up)** or **(A)** buttons **do not work**, you will need to use an alternate method, like Seedminer. For assistance with this matter, join Nintendo Homebrew on Discord and ask, in English, for help.

I do somewhat realize it's probably a much less significant edge case, but at the same time, we don't even mention any warning if either of those things are true, and the only mentions prior to that guide page to join the discord for general support (none of which even mention the possibility of needing a legacy method or the possibility of having a broken D-Pad/A button and are easily ignorable) are the following:
- The home page
- Get Started (model selection)
- Get Started (new 3ds page)

> GodMode9 also forces you to use d-pad for file unlocking by default, IIRC.

Actually no, at least on 1.9.2pre1, I can use the circle pad to unlock write permissions to, (normally wouldn't do this) for example, delete boot.firm.bak in CTRNand

lilyuwuu:
A general "dpad broken" warning can be put up if it actually becomes widespread -- dpad-up is only mentioned in the text you quoted for the purpose of safe mode.

Ultimately if people have physically broken buttons then they probably need to get help anyway because we need to guide them through whatever alternate method gets cooked up (using an alternate payload to install b9s, bypassing anything else that may require a dpad, etc)

_pages/en_US/installing-boot9strap-(pichaxx).txt
```diff
@@ -0,0 +1,69 @@
...
+### Required Reading
+
+This method of using Seedminer for further exploitation uses your `movable.sed` file to write a custom save file for Pokémon Picross, which can then be used with universal-otherapp to run SafeB9SInstaller.
+
+### What You Need
+
+* The game "Pokémon Picross" (free on eShop)
+* Your `movable.sed` file from completing [Seedminer](seedminer)
+* The latest release of [SafeB9SInstaller](https://github.com/d0k3/SafeB9SInstaller/releases/latest)
+* The latest release of [boot9strap](https://github.com/SciresM/boot9strap/releases/latest) *(standard boot9strap; not the `devkit` file, not the `ntr` file)*
```
MechanicalDragon0687, line 19: should update this to match the QOL change in cc2ff55 (boot9strap-1.3.zip)

PR #1830, opened 2020-12-29
reword installing boo9strap to say its the 3ds's sd

radostin04:
yes we got a complaint about this, im not a fan of "3DS's" either but it's the best we can do unless someone has a better idea


radostin04:
i changed it to console's hopefully this is better

Issue #1829, opened 2020-12-29
Some path notations may not be understood by some people.


mariohackandglitch:
**Pages with issue(s)**
https://3ds.hacks.guide/finalizing-setup

**Description of the issue(s)**
There are some steps that asks people to place files in the folder `xx/yy`. However, this notation is a bit technical and may not be understood by everyone. It would be better to say: place the files in the folder `yy` that is inside `xx`.

Here is a real example where an user got confused with this notation:
https://i.imgur.com/thrAcdR.png

So Section I, step 14 could be rewritten as:
Copy `GodMode9.firm` from the GodMode9 `.zip` to the `payloads` folder inside the `luma` folder on your SD card.

lilyuwuu:
Weirdly doesn't seem to be that common an issue from what I've seen on NH, but wouldn't necessarily be opposed to changing it. The problem is with that this is only feasible for places where you're only two folders deep -- what do we do when we have to direct someone to `/Nintendo 3DS/<ID0>/<ID1>/Nintendo DSiWare/`?

mariohackandglitch:
Maybe something in between would be using:  navigate to `xx` -> `yy`.

sjkim04:
Maybe we can use `sdmc:/`.

As an example, `sdmc:/luma/payloads/` would be less confusing.

sjkim04:
I’ll make a pull request on this, that would be more noticable.

PR #1831, opened 2020-12-30
finalizing-setup: add note if GM9 fails to start

ihaveamac:
I found someone who misspelled "payloads" as "playloads", causing
confusion for a while until I saw a directory listing. I don't know how
common exactly this is but I feel a note could be added here.

I'm not too happy about how this is worded though. Especially the path
which might need to be updated following #1829.


lilyuwuu:
"repeated attempts" is kind of vague. I'd suggest rewording the instructions to prevent any chance of holding START too late, and to declaratively say if it doesn't work to check file placement.


1. Power off your device
2. Hold the (Start) button
3. Power on your device while continuing to hold the (Start) button
4. If you were successful, you should boot into GodMode9
    - If you boot to the home menu, ensure that `GodMode9.firm` is in `luma -> payloads` and that `payloads` is properly spelled
5. If you are prompted to create an essential files backup, press (A) to do so, then press (A) to continue once it has completed

There's probably a better way to word this too.

MechanicalDragon0687:
I agree with adding the "Hold start" first, to prevent someone pressing start after power
I think paths should be consistent throughout the guide, or at least the page. Changing pattern to `luma -> payloads` is inconsistent with the current pattern.

lilyuwuu:
Changing pattern is a potential plan based on #1829. Should probably be a separate PR though since it'd change nearly every page.

PR #1832, opened 2020-12-30
Changed Legacy Method from Seedminer to Pichaxx

SketchMaster2001:
"If your (Right/Left Shoulder), (D-Pad Up) or (A) buttons do not work, you will need to use an alternate method, like Seedminer"

This statement is false, as you need those buttons to boot into safe-mode in Seedminer.

### DIFF
--- a/_pages/en_US/installing-boot9strap-(safecerthax).txt
+++ b/_pages/en_US/installing-boot9strap-(safecerthax).txt
@@ -8,7 +8,7 @@ title: "Installing boot9strap (safecerthax)"
 
 safecerthax is compatible with all Old 3DS and Old 2DS devices in all regions.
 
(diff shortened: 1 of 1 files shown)


techmuse8:
Well for pichaxx you would need seedminer (which doesn’t need LR) to get
your movable to do pichaxx.

SketchMaster2001:
yes, but seedminer will redirect you to boot9strap (USM) which needs the shoulder buttons, while pichaxx will redirect you to boot9strap (Frogtool) which doesn't need the shoulder buttons

daeny99:
And if you scroll down it says "Legacy Methods" under the seedminer route which when clicked will tell you to use Seedminer + BannerBomb3 + Fredtool if your shoulder buttons do not work.

SketchMaster2001:
alright then

PR #1845, opened 2021-02-18
add universal updater to the finalizing setup

MechanicalDragon0687:
Every other console has a hb app store, lets add one to 3ds too

Adds Universal Updater to the finalizing setup page for 3DS


MechanicalDragon0687:
on 48 hour hold for input

PR #1850, opened 2021-03-01
Update finalizing settup page (and add troubleshooting section) to avoid unnecessary CTR transfers on section II

TheGoofBall64:
updates finalizing setup page section II to link to a (new) troubleshooting section to avoid unnecessary CTRtransfers when your device fails to update

adds troubleshooting section on the troubleshooting page.
changes finalizing setup page to link to this troubleshooting section.

### DIFF
--- a/_pages/en_US/finalizing-setup.txt
+++ b/_pages/en_US/finalizing-setup.txt
@@ -67,7 +67,7 @@ If, before following this guide, you already had an EmuNAND setup and would like
   + Updates while using B9S + Luma (what you have) are safe
   + The updater may display a message saying "Your system is up to date" instead of updating. This is normal if you are already up to date; continue with the next section
   + If this gives you an error, set your DNS settings to "auto"
-  + If this still gives you an error, [follow this troubleshooting guide](troubleshooting#wifi-connection-errors-after-installing-cfw-to-your-3ds), then try updating again
+  + If this still gives you an error, [follow this troubleshooting guide](troubleshooting#wifi-connection-errors-after-installing-cfw), then try updating again
 
 #### Section III - Homebrew Launcher
 
--- a/_pages/en_US/finalizing-setup.txt
+++ b/_pages/en_US/finalizing-setup.txt
@@ -67,7 +67,7 @@ If, before following this guide, you already had an EmuNAND setup and would like
   + Updates while using B9S + Luma (what you have) are safe
   + The updater may display a message saying "Your system is up to date" instead of updating. This is normal if you are already up to date; continue with the next section
   + If this gives you an error, set your DNS settings to "auto"
-  + If this still gives you an error. Try to get closer to your WiFi, restart your WiFi router (unplug for 10 seconds, plug back in), restart your 3ds, disable your fire wall, connect to a different wifi (such as a mobile hotspot)
+  + If this still gives you an error. Try to get closer to your WiFi router, restart your WiFi router (unplug for 10 seconds, plug back in), restart your 3DS, disable your fire wall, connect to a different WiFi (such as a mobile hotspot)
   + If this still gives you an error, [follow CTRTransfer](ctrtransfer), then try updating again
 
 #### Section III - Homebrew Launcher
--- a/_pages/en_US/finalizing-setup.txt
+++ b/_pages/en_US/finalizing-setup.txt
@@ -67,6 +67,7 @@ If, before following this guide, you already had an EmuNAND setup and would like
   + Updates while using B9S + Luma (what you have) are safe
   + The updater may display a message saying "Your system is up to date" instead of updating. This is normal if you are already up to date; continue with the next section
   + If this gives you an error, set your DNS settings to "auto"
+  + If this still gives you an error. Try to get closer to your WiFi, restart your WiFi router (unplug for 10 seconds, plug back in), restart your 3ds, disable your fire wall, connect to a different wifi (such as a mobile hotspot)
   + If this still gives you an error, [follow CTRTransfer](ctrtransfer), then try updating again
 
 #### Section III - Homebrew Launcher
--- a/_pages/en_US/finalizing-setup.txt
+++ b/_pages/en_US/finalizing-setup.txt
@@ -67,7 +67,7 @@ If, before following this guide, you already had an EmuNAND setup and would like
   + Updates while using B9S + Luma (what you have) are safe
   + The updater may display a message saying "Your system is up to date" instead of updating. This is normal if you are already up to date; continue with the next section
   + If this gives you an error, set your DNS settings to "auto"
-  + If this still gives you an error. Try to get closer to your WiFi router, restart your WiFi router (unplug for 10 seconds, plug back in), restart your 3DS, disable your fire wall, connect to a different WiFi (such as a mobile hotspot)
+  + If this still gives you an error, Try to get closer to your WiFi router, restart your WiFi router (unplug for 10 seconds, plug back in), restart your 3DS, disable your fire wall, connect to a different WiFi (such as a mobile hotspot)
   + If this still gives you an error, [follow CTRTransfer](ctrtransfer), then try updating again
 
 #### Section III - Homebrew Launcher
--- a/_pages/en_US/finalizing-setup.txt
+++ b/_pages/en_US/finalizing-setup.txt
@@ -67,8 +67,7 @@ If, before following this guide, you already had an EmuNAND setup and would like
   + Updates while using B9S + Luma (what you have) are safe
   + The updater may display a message saying "Your system is up to date" instead of updating. This is normal if you are already up to date; continue with the next section
   + If this gives you an error, set your DNS settings to "auto"
-  + If this still gives you an error, Try to get closer to your WiFi router, restart your WiFi router (unplug for 10 seconds, plug back in), restart your 3DS, disable your fire wall, connect to a different WiFi (such as a mobile hotspot)
-  + If this still gives you an error, [follow CTRTransfer](ctrtransfer), then try updating again
+  + If this still gives you an error, [follow this troubleshooting guide](troubleshooting#wifi-connection-errors-after-installing-cfw-to-your-3ds), then try updating again
 
 #### Section III - Homebrew Launcher
 
--- a/_pages/en_US/finalizing-setup.txt
+++ b/_pages/en_US/finalizing-setup.txt
@@ -67,7 +67,7 @@ If, before following this guide, you already had an EmuNAND setup and would like
   + Updates while using B9S + Luma (what you have) are safe
   + The updater may display a message saying "Your system is up to date" instead of updating. This is normal if you are already up to date; continue with the next section
   + If this gives you an error, set your DNS settings to "auto"
-  + If this still gives you an error, [follow this troubleshooting guide](troubleshooting#wifi-connection-errors-after-installing-cfw), then try updating again
+  + If this still gives you an error, [follow this troubleshooting guide](troubleshooting#system-update-error-after-installing-cfw), then try updating again
 
 #### Section III - Homebrew Launcher
 
--- a/_pages/en_US/finalizing-setup.txt
+++ b/_pages/en_US/finalizing-setup.txt
@@ -67,7 +67,7 @@ If, before following this guide, you already had an EmuNAND setup and would like
   + Updates while using B9S + Luma (what you have) are safe
   + The updater may display a message saying "Your system is up to date" instead of updating. This is normal if you are already up to date; continue with the next section
   + If this gives you an error, set your DNS settings to "auto"
-  + If this still gives you an error, [follow this troubleshooting guide](troubleshooting#system-update-error-after-installing-cfw), then try updating again
+  + If this still gives you an error, [follow this troubleshooting guide](troubleshooting#system-update-error-after-installing-cfw)
 
 #### Section III - Homebrew Launcher
 
--- a/_pages/en_US/troubleshooting.txt
+++ b/_pages/en_US/troubleshooting.txt
@@ -166,5 +166,5 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
 1. reboot your WiFi router
 1. conect to a different WiFi conection (mobile hotspot, etc.)
 1. Nintendo servers may be down. Try again later
-1. If none of the suggestions above worked, [follow CTRTransfer](ctrtransfer), then try again
+1. If none of the options above worked, [follow CTRTransfer](ctrtransfer), then try again
 1. For support (in English), join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp)
--- a/_pages/en_US/troubleshooting.txt
+++ b/_pages/en_US/troubleshooting.txt
@@ -158,10 +158,12 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
 
 ## WiFi connection errors after installing CFW
 
-1. You get an error while trying to use WiFi (during system update, loading the Browser/Eshop, ect.)
+1. You get an error while trying to use WiFi (during system update, loading the browser/Eshop, ect.)
 1. Try moving closer to your WiFi router, and rebooting your 3DS
 1. Try setting your DNS settings to "auto"
 1. Try deleting your WiFi connections, and signing into your Wifi again
 1. Try rebooting your WiFi router
 1. Try conecting to a different WiFi conection (mobile hotspot, ect.)
+1. Try again later (Nintendo servers are down, ect.)
 1. If this still gives you an error, [follow CTRTransfer](ctrtransfer), then try again
+1. For support (in English), join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp)
--- a/_pages/en_US/troubleshooting.txt
+++ b/_pages/en_US/troubleshooting.txt
@@ -166,5 +166,5 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
 1. Reboot your WiFi router
 1. Connect to a different WiFi connection, like a mobile hotspot
 1. Nintendo servers may be down. Try again later
-1. If none of the steps above worked, [follow CTRTransfer](ctrtransfer), then try updating again
+1. If you still get an error, [follow CTRTransfer](ctrtransfer), then try again
 1. For support (in English), join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp)
--- a/_pages/en_US/troubleshooting.txt
+++ b/_pages/en_US/troubleshooting.txt
@@ -161,7 +161,7 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
 1. You get an error while trying to use WiFi (During system update, loading the Browser/Eshop, ect.)
 1. Try moving closer to your WiFi router, and rebooting your 3DS
 1. Try setting your DNS settings to "auto"
-1. Try deleting your WiFi connection, and signing into your Wifi network again
+1. Try deleting your WiFi connections, and signing into your Wifi again
 1. Try rebooting your WiFi router
 1. Try conecting to a different WiFi conection (Mobile hotspot, ect.)
-1. If this still gives you an error, [follow CTRTransfer](ctrtransfer), then try updating again
+1. If this still gives you an error, [follow CTRTransfer](ctrtransfer), then try again
--- a/_pages/en_US/troubleshooting.txt
+++ b/_pages/en_US/troubleshooting.txt
@@ -158,8 +158,7 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
 
 ## System update error after installing CFW
 
-  + You get "An error has occured" during system update
-  + Reboot your 3DS after each step and try updating again
+  + Occasionally, updates may fail to install after installing CFW. To fix this, reboot your 3DS after each step of this section, then try updating again.
 1. Set your DNS settings to "Auto"
 1. Move closer to your WiFi router
 1. Update from Safe Mode by turning off the console, holding L+R+Up+A on boot, and following the prompts
--- a/_pages/en_US/troubleshooting.txt
+++ b/_pages/en_US/troubleshooting.txt
@@ -158,7 +158,7 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
 
 ## System update error after installing CFW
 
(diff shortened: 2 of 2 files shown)


win9x-se:
a few things

Unplugging routers can cause a ton of problems, i.e. needing to reflash or otherwise fix it, not all routers are made to be unplugged on whim

Not everyone has access to their router/internet system and thus can't do things like disable router wide firewall, etc

The 3DS itself doesn't have a firewall, so that wouldn't apply

Most of these steps are redundant as a lot of the time, it comes down to "bring your console closer, and reboot".

MechanicalDragon0687:
the more recent issues we have had with people updating is due to their wifi connection settings still being set for the dns hacks or being broken. I have had my connection fail even after clearing the dns settings and had to recreate it completely to fix it.

Perhaps adding a Wifi Troubleshooting section on the troubleshooting page would make more sense considering there are several methods that require wifi connection manipulation now?

TheGoofBall64:
Yeah. Your probably right. I'll add something like that

TheGoofBall64:
> Unplugging routers can cause a ton of problems, i.e. needing to reflash or otherwise fix it, not all routers are made to be unplugged on whim

Yeah your right. I'll fix that

> Not everyone has access to their router/internet system and thus can't do things like disable router wide firewall, etc

This is kind of a redundant argument. Yes it is almost never the problem, but If you are having problems and have access to, and the know how, to disable your firewall, go ahead and try. it shouldn't matter if some people can not access it. It's still a potential problem.

> The 3DS itself doesn't have a firewall, so that wouldn't apply

Yes, the 3ds doesn't have any sort of firewall. But your WiFi itself could have a firewall, and could block specific things your 3ds calls for. As per Nintendo points out in their own network troubleshooting guide

> Most of these steps are redundant as a lot of the time, it comes down to "bring your console closer, and reboot".

Yes. You are right. But again, I myself has had to restart my WiFi network multiple times. Even after going closer and rebooting. And even had to try another network entirely because my WiFi just didn't work that day. This is why (I'm going to) adding a troubleshooting section on there like Randal suggested would be a good idea. Most of the time it does come down to "bring your console closer, and reboot". But it still happens

MechanicalDragon0687:
adding a "check the firewall settings in your router" is more likely to confuse the common user of the guide. If they have firewall settings in the router blocking things, and they set them up, it is likely unnecessary to tell them to check them. if they have firewall settings in the router and they didnt, telling them to check it is moot

TheGoofBall64:
Ok. I went add ''check firewall" then

TheGoofBall64:
If you have any other suggestions on what I should add to the troubleshooting section please go ahead. Thank you

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,15 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
 1. Your device is bricked
 1. You will need to get an ntrboot-comptible flashcart (one of the ones on [this list](ntrboot) or a [hardmod](https://gbatemp.net/threads/414498/), or repair / replace your device
 1. For support (in English), join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp)
+
+## WiFi connection errors after installing CFW
+
+1. You get an error while trying to use WiFi (during system update, loading the browser/Eshop, ect.)
```
lilyuwuu, line 161: This is a statement. It shouldn't be a 'step' in the troubleshooting.

lilyuwuu: `Eshop` -> `eShop`; `ect.` -> `etc.`

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,15 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
...
+1. Try moving closer to your WiFi router, and rebooting your 3DS
+1. Try setting your DNS settings to "auto"
```
lilyuwuu, line 163: "Auto" should be capitalized

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,15 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
...
+1. Try rebooting your WiFi router
+1. Try deleting your WiFi connection/s, then sign into your Wifi again
```
lilyuwuu, line 165: Use consistent capitalization of WiFi

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,15 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
...
+1. Try conecting to a different WiFi conection (mobile hotspot, ect.)
```
lilyuwuu, line 166: `ect.` -> `etc.`

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,15 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
 1. Your device is bricked
 1. You will need to get an ntrboot-comptible flashcart (one of the ones on [this list](ntrboot) or a [hardmod](https://gbatemp.net/threads/414498/), or repair / replace your device
 1. For support (in English), join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp)
+
+## WiFi connection errors after installing CFW
+
+1. You get an error while trying to use WiFi (during system update, loading the browser/Eshop, ect.)
+1. Try moving closer to your WiFi router, and rebooting your 3DS
```
lilyuwuu, line 162: Remove the comma

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,15 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
...
+1. Try setting your DNS settings to "auto"
+1. Try rebooting your WiFi router
+1. Try deleting your WiFi connection/s, then sign into your Wifi again
+1. Try conecting to a different WiFi conection (mobile hotspot, ect.)
+1. Try again later (Nintendo servers are down, ect.)
```
lilyuwuu, line 167: This doesn't really work as an `etc.` If your only example is Nintendo servers being down, use `Nintendo servers may be down. Try again later` or something.

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,15 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
...
+1. If this still gives you an error, [follow CTRTransfer](ctrtransfer), then try again
```
lilyuwuu, line 168: "this" is ambiguous in this context; needs to be reworded

_pages/en_US/troubleshooting.txt
lilyuwuu, line 165: `Sign in` implies that there's a portal/login page, which the 3DS doesn't support. Use `connect to` instead.

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,15 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
...
+1. For support (in English), join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp)
```
lilyuwuu, line 169: The order of this steps should be from easiest to hardest to do, which would be (IMO):

1. Set DNS to Auto
2. Move closer to router
3. Delete and re-add WiFi network
4. Try a different WiFi connection
5. Nintendo's servers might be down
6. CTRTransfer
7. Ask for help

_pages/en_US/finalizing-setup.txt
```diff
@@ -67,7 +67,7 @@ If, before following this guide, you already had an EmuNAND setup and would like
   + Updates while using B9S + Luma (what you have) are safe
   + The updater may display a message saying "Your system is up to date" instead of updating. This is normal if you are already up to date; continue with the next section
   + If this gives you an error, set your DNS settings to "auto"
-  + If this still gives you an error, [follow CTRTransfer](ctrtransfer), then try updating again
+  + If this still gives you an error, [follow this troubleshooting guide](troubleshooting#wifi-connection-errors-after-installing-cfw), then try updating again
```
lilyuwuu, line 70: You are suggesting multiple ideas, each of which should be tried before updating again. This isn't necessarily clear from the text of your troubleshooting guide.

_pages/en_US/troubleshooting.txt
lilyuwuu, line 167: `ect.` -> `etc.`

_pages/en_US/troubleshooting.txt
lilyuwuu, line 169: I don't think `Try` should be at the beginning of every sentence (use the form used with CTRTransfer, which is `then try again`). It's left unspoken what to do after you reboot you WiFi router/set DNS to auto/whatever, which is unclear.

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,15 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
 1. Your device is bricked
 1. You will need to get an ntrboot-comptible flashcart (one of the ones on [this list](ntrboot) or a [hardmod](https://gbatemp.net/threads/414498/), or repair / replace your device
 1. For support (in English), join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp)
+
+## WiFi connection errors after installing CFW
```
lilyuwuu, line 159: This is a bit vague. The only case we're really worried about is system updating, as Nintendo doesn't offer any error codes when something goes wrong during system updates. If an error occurs in the browser or in eShop, there is usually an error code that goes along with it. We're not dealing with browser or eShop connection issues in this portion of the guide, so I don't think it should be so vague.

This also means an online services ban could be misinterpreted as a WiFi connection error.

_pages/en_US/troubleshooting.txt
lilyuwuu, line 165: A user should only have one WiFi connection that they should be connecting to at this moment. I don't see the need to delete *multiple* connections.

lilyuwuu (changes requested):

TheGoofBall64 (commented):
fixed.

but I never put in to try and hold `A+R+L+Up(DPad) on boot (to attempt an update in recovery mode)` anywhere. and I'm having trouble on how to implement it. because the ideal situation is to try this on every option after "move closer to your WiFi". do you have any ideas on how to implement this

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,16 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
...
+
+  + You get an error while trying to use WiFi (during system update)
+  + reboot your 3DS after each option and try updating again
+1. set your DNS settings to "Auto"
+1. move closer to your WiFi router
+1. delete your WiFi connection, then connect to your WiFi again
+1. reboot your WiFi router
+1. conect to a different WiFi conection (mobile hotspot, etc.)
+1. Nintendo servers may be down. Try again later
+1. If none of the options above worked, [follow CTRTransfer](ctrtransfer), then try updating again
```
lilyuwuu, line 169: The beginning of each point should be capitalized.

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,16 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
...
+
+## System update error after installing CFW
+
+  + You get an error while trying to use WiFi (during system update)
+  + reboot your 3DS after each option and try updating again
+1. set your DNS settings to "Auto"
+1. move closer to your WiFi router
+1. delete your WiFi connection, then connect to your WiFi again
+1. reboot your WiFi router
+1. conect to a different WiFi conection (mobile hotspot, etc.)
```
lilyuwuu, line 167: "Connect" is misspelled, as is "connection". 

"Connect to a different WiFi connection" is a bit repetitive, but the 3DS calls them 'connections' so it *should* be fine...

It's a bit awkward to use `etc.` when you only have one item as an example. If the only 'different WiFi connection' you can think of is a mobile hotspot, it might be better to just word it as "Connect to a different WiFi connection, like a mobile hotspot [or something else]"

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,16 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
 1. Your device is bricked
 1. You will need to get an ntrboot-comptible flashcart (one of the ones on [this list](ntrboot) or a [hardmod](https://gbatemp.net/threads/414498/), or repair / replace your device
 1. For support (in English), join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp)
+
+## System update error after installing CFW
+
+  + You get an error while trying to use WiFi (during system update)
```
lilyuwuu, line 161: An error when trying to use WiFi during system update, as opposed to trying to use something else?...

The best way to clarify this section would be to show that the error that people should see is "An error has occurred".

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,16 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
...
+  + reboot your 3DS after each option and try updating again
```
lilyuwuu, line 162: option -> step; if it's an option, then people can follow them out of order (though they likely wouldn't)

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,16 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
...
+1. set your DNS settings to "Auto"
+1. move closer to your WiFi router
+1. delete your WiFi connection, then connect to your WiFi again
```
lilyuwuu, line 165: A bit awkwardly worded IMO. 'm partial to using the term `reconnect` but it might be unnecessarily vague.

lilyuwuu (changes requested):

lilyuwuu:
A Safe Mode step could just be
`1. Update from Safe Mode by turning off the console, holding L+R+Up+A on boot, and following the prompts` or something

_pages/en_US/troubleshooting.txt
TheGoofBall64: oh god. how did i miss that?

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,17 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
 1. Your device is bricked
 1. You will need to get an ntrboot-comptible flashcart (one of the ones on [this list](ntrboot) or a [hardmod](https://gbatemp.net/threads/414498/), or repair / replace your device
 1. For support (in English), join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp)
+
+## System update error after installing CFW
+
+  + You get "An error has occured" during system update
+  + Reboot your 3DS after each step and try updating again
```
lilyuwuu, line 162: Maybe replace these two bullet points with a paragraph that reads

"Occasionally, updates may fail to install after installing CFW. To fix this, reboot your 3DS after each step of this section, then try updating again."

This would be in line with other troubleshooting steps that add more information. 

Alternatively, since this is currently only linked on Finalizing Setup, you could just have " Reboot your 3DS after each step, then try updating again" as one sentence.

Third idea: Don't have any text before the steps, then add "then try updating again" after each step.

lilyuwuu (commented):
The rest generally looks good to me!

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,16 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
 1. Your device is bricked
 1. You will need to get an ntrboot-comptible flashcart (one of the ones on [this list](ntrboot) or a [hardmod](https://gbatemp.net/threads/414498/), or repair / replace your device
 1. For support (in English), join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp)
+
+## System update error after installing CFW
+
+  + Occasionally, updates may fail to install after installing CFW. To fix this, reboot your 3DS after each step of this section, then try updating again.
```
lilyuwuu, line 161: This shouldn't be a bullet point/indented.

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,16 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
...
+1. Set your DNS settings to "Auto"
+1. Move closer to your WiFi router
+1. Update from Safe Mode by turning off the console, holding L+R+Up+A on boot, and following the prompts
+1. Delete your WiFi connection, then reconnect to your WiFi again
+1. Reboot your WiFi router
+1. Connect to a different WiFi connection, like a mobile hotspot
+1. Nintendo servers may be down. Try again later
+1. If none of the steps above worked, [follow CTRTransfer](ctrtransfer), then try updating again
```
lilyuwuu, line 169: Is this intentional, as some way to discourage CTRTransfering? It's inconsistent to the rest of the text.

Also, "then try updating again" shouldn't be necessary because it's said at the beginning of this section.

TheGoofBall64: no it wasn't intentional. sorry, ill fix that

_pages/en_US/troubleshooting.txt
TheGoofBall64: fixed

lilyuwuu (approved):
Looks good

_pages/en_US/troubleshooting.txt
```diff
@@ -162,7 +162,7 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
 
 1. Set your DNS settings to "Auto"
 1. Move closer to your WiFi router
-1. Update from Safe Mode by turning off the console, holding L+R+Up+A on boot, and following the prompts
+1. Update from Safe Mode by turning off the console, holding (A) + (Right Shoulder) + (Left Shoulder) + (D-Pad Up) on boot, and following the on screen prompts
```
lilyuwuu, line 165: A bit picky, but the guide uses the order of `(Left Shoulder) + (Right Shoulder) + (D-Pad Up) + (A)` to describe the buttons.

lilyuwuu, line 165: There should be a hyphen between on and screen (on-screen).

TheGoofBall64:
when will this get merged? It's been 6 days since any significant changes were made. and it was approved

lilyuwuu:
No one else has looked at it so I'm a bit hesitant to merge it. The policy is usually to get at least a couple people to look before merging. That's also why if you look at the PR history a lot of the changes that weren't to fix a typo or something smaller took a while to merge.

That being said, nothing will get done without a strict time limit so I'll go ahead and merge this in about 48 hours if no one else has any thoughts/opposition.

win9x-se:
troubleshooting seems to have an error with the numbering
image

MechanicalDragon0687:
> troubleshooting seems to have an error with the numbering
> image

this is normal. the numbering isnt hardcoded until built.

win9x-se:
> 
> 
> > troubleshooting seems to have an error with the numbering
> > image
> 
> this is normal. the numbering isnt hardcoded until built.

yeah i just realized that, sorry

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,17 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
 1. Your device is bricked
 1. You will need to get an ntrboot-comptible flashcart (one of the ones on [this list](ntrboot) or a [hardmod](https://gbatemp.net/threads/414498/), or repair / replace your device
 1. For support (in English), join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp)
+
+## System update error after installing CFW
+
+ Occasionally, updates may fail to install after installing CFW. To fix this, reboot your 3DS after each step of this section, then try updating again.
```
GlaZedBelmont, line 161: We usually use "console" or "device" when talking about the 3DS, the rest of your edits don't say 3DS

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,17 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
...
+
+1. Set your DNS settings to "Auto"
+1. Move closer to your WiFi router
+1. Update from Safe Mode by turning off the console, holding (Left Shoulder) + (Right Shoulder) + (D-Pad Up) + (A) on boot, and following the on-screen prompts
+1. Delete your WiFi connection, then reconnect to your WiFi again
+1. Reboot your WiFi router
+1. Connect to a different WiFi connection, like a mobile hotspot
+1. Nintendo servers may be down. Try again later
```
GlaZedBelmont, line 169: small nitpick but why do you say "[...] may be down. Try again later" then "[...], then try again" maybe using a comma on the former would be better

_pages/en_US/troubleshooting.txt
```diff
@@ -155,3 +155,17 @@ Browser based exploits (such as browserhax or 2xrsa) are often unstable and cras
...
 1. You will need to get an ntrboot-comptible flashcart (one of the ones on [this list](ntrboot) or a [hardmod](https://gbatemp.net/threads/414498/), or repair / replace your device
 1. For support (in English), join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp)
+
+## System update error after installing CFW
+
+ Occasionally, updates may fail to install after installing CFW. To fix this, reboot your 3DS after each step of this section, then try updating again.
+
+1. Set your DNS settings to "Auto"
+1. Move closer to your WiFi router
+1. Update from Safe Mode by turning off the console, holding (Left Shoulder) + (Right Shoulder) + (D-Pad Up) + (A) on boot, and following the on-screen prompts
```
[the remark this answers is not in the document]
lilyuwuu: That is incorrect -- in English, verb tenses should be consistent. The first instance is "turning" (off the console), so the next ones should indeed be "holding" and "following". "Update" is separate.

_pages/en_US/troubleshooting.txt
lilyuwuu: It'd technically be grammatically incorrect because "try again later" is a full sentence. You can use a semicolon to separate them though.

_pages/en_US/troubleshooting.txt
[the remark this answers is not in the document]
TheGoofBall64: I speak English as my main language. and even I never noticed this. I automatically write it with the correct verb tenses and everything

PR #1860, opened 2021-03-21
add Japanese to the language list

MechanicalDragon0687:
looks like the guide's japanese translation is getting translations through pretty quickly.
Jpn guide is about 95% finished translating which is actually quite a bit more than some of the other languages we have up

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -156,7 +156,7 @@ exclude: [
   # _pages/hu_HU,
   _pages/id_ID,
   # _pages/it_IT,
-  #_pages/ja_JP,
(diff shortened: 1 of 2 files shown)


_config.yml
```diff
@@ -156,7 +156,7 @@ exclude: [
   # _pages/hu_HU,
   _pages/id_ID,
   # _pages/it_IT,
-  _pages/ja_JP,
+  #_pages/ja_JP,
```
lilyuwuu, line 159: There's a space missing here, not sure if that makes a difference though.

PR #1863, opened 2021-03-24
"private" exists.

radostin04:
the "private" folder stores some data for certain apps and it may be present on a stock system, so it doesn't exactly make sense to delete it.


radostin04:
also fixed a typo in uninstall cfw thanks for telling mr. burgers

PR #1871, opened 2021-04-05
Easing up dumping boot9, boot11 & OTP

ayechief:
**Description**

### DIFF
--- a/_pages/en_US/finalizing-setup.txt
+++ b/_pages/en_US/finalizing-setup.txt
@@ -138,10 +138,11 @@ If, before following this guide, you already had an EmuNAND setup and would like
 1. Select "Dump OTP.bin"
 1. Press (A) to confirm
 1. Press (A) to continue
+1. Press (B) to return to the main menu
 1. Press (Home) to bring up the action menu
 1. Select "Poweroff system" to power off your device
 1. Insert your SD card into your computer
-1. Copy `<date>_<serialnumber>_sysnand_###.bin`, `<date>_<serialnumber>_sysnand_###.bin.sha`, `essential.exefs`, and `boot9.bin` from the `/gm9/out/` folder on your SD card to a safe location on your computer
+1. Copy `<date>_<serialnumber>_sysnand_###.bin`, `<date>_<serialnumber>_sysnand_###.bin.sha`, `essential.exefs`, `boot9.bin`, 'boot11.bin' and 'OTP.bin' from the `/gm9/out/` folder on your SD card to a safe location on your computer
   + Make backups in multiple locations (such as online file storage)
   + These backups will save you from a brick and/or help you recover files from the NAND image if anything goes wrong in the future
 1. Delete `<date>_<serialnumber>_sysnand_###.bin` and `<date>_<serialnumber>_sysnand_###.bin.sha` from the `/gm9/out/` folder on your SD card after copying it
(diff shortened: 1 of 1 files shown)


_pages/en_US/finalizing-setup.txt
```diff
@@ -131,17 +131,18 @@ If, before following this guide, you already had an EmuNAND setup and would like
...
-1. Select "Exit"
-1. Press (A) to relock write permissions if prompted
-1. Navigate to `[M:] MEMORY VIRTUAL`
-1. Press (A) on `boot9.bin` to select it
-1. Select "Copy to 0:/gm9/out"
+1. Select "Dump Options"
+1. Select "Dump Boot9.bin & Boot11.bin"
+1. Press (A) to confirm
+1. Press (A) to continue
+1. Select "Dump OTP.bin"
```
hansoljin, line 138: This step is unnecessary as `essential.exefs` includes `OTP.bin` and is dumped much earlier in the guide.
Perhaps consider replacing "Dump OTP.bin" with "Dump essential.exefs" instead, to make sure it is actually dumped? (We do see a handful of helpees in the Discord asking about no exefs) But someone else should check on that first.

_pages/en_US/finalizing-setup.txt
```diff
@@ -131,17 +131,18 @@ If, before following this guide, you already had an EmuNAND setup and would like
   + This process will take some time
   + If you get an error, make sure that you have at least 1.3GB of free space on your SD card
 1. Press (A) to continue
-1. Press (B) to return to the main menu
-1. Select "Exit"
-1. Press (A) to relock write permissions if prompted
```
hansoljin, line 136: Please double-check where this line can be placed. It still prompts to relock write permissions when exiting to the main screen as the perms were unlocked while setting up Luma3DS to CTRNAND; however, this relocking step is removed here and is not referenced again.

An easy way to do this is to actually follow along the process yourself, documenting every single step you take.

_pages/en_US/finalizing-setup.txt
```diff
@@ -131,17 +131,18 @@ If, before following this guide, you already had an EmuNAND setup and would like
...
+1. Press (A) to continue
+1. Select "Dump OTP.bin"
+1. Press (A) to confirm
 1. Press (A) to continue
+1. Press (B) to return to the main menu
 1. Press (Home) to bring up the action menu
 1. Select "Poweroff system" to power off your device
 1. Insert your SD card into your computer
-1. Copy `<date>_<serialnumber>_sysnand_###.bin`, `<date>_<serialnumber>_sysnand_###.bin.sha`, `essential.exefs`, and `boot9.bin` from the `/gm9/out/` folder on your SD card to a safe location on your computer
+1. Copy `<date>_<serialnumber>_sysnand_###.bin`, `<date>_<serialnumber>_sysnand_###.bin.sha`, `essential.exefs`, `boot9.bin`, 'boot11.bin' and 'OTP.bin' from the `/gm9/out/` folder on your SD card to a safe location on your computer
```
hansoljin, line 145: Same as above, `OTP.bin` is unnecessary.

As well, rename `boot9.bin` and `boot11.bin` to include `<serialnumber>_` in front, so that it is consistent with what is actually being dumped, as well being consistent with the guide's formatting

hansoljin (changes requested):

Issue #1864, opened 2021-03-26
Old browser exploit isn't possible on N2DSLL (JPN) on 11.6


malvarenga123:
**Pages with issue(s)**

https://3ds.hacks.guide/get-started-(new-3ds)

**Description of the issue(s)**
The guide indicates that for a New 2DS XL (LL) at firmware 11.6 one should use SSLoth-Browser exploit.
The issue is that the browser is unusable on firmware 11.6. If you try to access any page using the browser, a message saying you should update the software pops up.
Similar to what is reported here: https://gbatemp.net/threads/stumped-on-installing-cfw-on-japanese-n2dsll.580973/
Error 032-0420 happens if you use the spoofed DNS before entering the browser at least once.

The only solution, which was also mentioned in that thread, was to update to version 11.14 and then run new-browserhax-xl.

lilyuwuu:
Aware of the issue, working on a way to resolve it. We've thrown around the following ideas so far:

- add warning about jpn n3ds under required reading, where there is currently a warning about cart updated consoles
- add warning about jpn n3ds within Section III - Launching SafeB9SInstaller, but we'd need to reword it since it just says "if you get an error, follow this troubleshooting guide"
- remove the method entirely for being more trouble than it's worth, and just tell people to update their consoles

malvarenga123:
My vote goes to
> remove the method entirely for being more trouble than it's worth, and just tell people to update their consoles
As long as 11.14 remains the latest update, it's just easier to update and run new-browserhax-xl.

MechanicalDragon0687:
moot point now

PR #1874, opened 2021-04-10
Implement old-browserhax-xl

lilyuwuu:
Implements old-browserhax-xl. **Affects main guide path**.

This PR...
- Reworks `Installing boot9strap (Browser)` to be compatible with *both* Old 3DS and New 3DS. 
- Unsplits guide path (just one page now, "get-started"). 

Other changes in this PR:
- Removed SSLoth-Browser (#1864), replaced with "update your console"


hansoljin:
old-browserhax-xl is utilizing https://zoogie.github.io/web/nbhax/ and not obhax-xl. Should be amended to reflect that. 

I'm unsure whether the QR code needs to be updated, as I only tested the direct link.

All regions are now supported according to zoogie. safecerthax is no longer necessary.

PR #1886, opened 2021-05-26
Add kramdown-parser-gfm to Gemfile

lifehackerhansol:
**Description**

This PR adds `kramdown-parser-gfm` to Gemfile. This doesn't really have anything to do with the page, but for local testing, it keeps failing to build the site without it, and it isn't really messing with the site either way, so why not? 

Admittedly this is more of an annoyance thing than anything else, but I assume other people are having this issue as well when offline testing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,4 @@
 source "https://rubygems.org"
 gem 'jekyll-redirect-from'
+gem 'kramdown-parser-gfm'
 gemspec


NightYoshi370 (approved):
I needed to do this on my local copy. Glad it's being reflected in a PR

lifehackerhansol:
Force-push was to change commit author. Nothing else changed.

PR #1889, opened 2021-06-04
Update link so that it displays properly in Spanish

ManiacOfGitHub:
Update link so that it displays properly on the "Install boot9strap (Browser)" page for the Spanish guide as it would not show as a link.


lifehackerhansol:
This should be fixed on Crowdin instead as this is a translation error.

Please visit https://crowdin.com/project/3ds-guide and resolve the issue as there will be conflicts if fixed from GitHub's side.

PR #1890, opened 2021-06-10
add dumping essential.exefs to finalize

lifehackerhansol:
**Description**

By request of @lilyuwuu `essential.exefs` is now dumped during Finalizing Setup. `.noexefs` is no more.

GM9Megascript has everything.

### DIFF
--- a/_pages/en_US/finalizing-setup.txt
+++ b/_pages/en_US/finalizing-setup.txt
@@ -136,6 +136,9 @@ If, before following this guide, you already had an EmuNAND setup and would like
 1. Select "Dump Boot9.bin & Boot11.bin"
 1. When prompted, press (A) to proceed
 1. Press (A) to continue
+1. Select "Dump essential.exefs"
+1. When prompted, press (A) to proceed
+1. Press (A) to continue
 1. Press (B) to return to the main menu
 1. Select "Exit"
 1. Press (A) to relock write permissions if prompted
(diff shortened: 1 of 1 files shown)


lifehackerhansol:
If we do end up going the long way for that, you may as well dump boot9 from MEMORY VIRTUAL as well. It wouldn't make sense to visit "Dump options" once to do that.

I guess I'll revert that commit here.

_pages/en_US/finalizing-setup.txt
```diff
@@ -139,6 +139,11 @@ If, before following this guide, you already had an EmuNAND setup and would like
 1. Press (B) to return to the main menu
 1. Select "Exit"
 1. Press (A) to relock write permissions if prompted
+1. Navigate to `[S:] SYSNAND VIRTUAL`
+1. Press (A) on `essential.exefs` to select it
+1. Select "Copy to 0:/gm9/out"
+  + If prompted that "Destination already exists", press (A) on "Overwrite file(s)"
```
lilyuwuu, line 145: Maybe use `If you see "Destination already exists"`

lilyuwuu (approved):
lgtm

_pages/en_US/finalizing-setup.txt
lifehackerhansol: Amended.

PR #1894, opened 2021-07-14
Replace cfdisk with mkfs.fat in Formatting SD (Linux)

lifehackerhansol:
**Description**

The current `cfdisk` instructions are somewhat destructive, and doesn't actually set up cluster sizes in the way that is needed specifically for the guide.

With that in mind, it was replaced to `mkfs.fat` instead, which is simple, shorter, and more accurate for the guide's needs.


lifehackerhansol:
An idea I had for this pull request was to just overhaul all the formatting pages. They work, they could just be better, which was the point of this PR anyway. Notably, the usage of the official SD Formatter for Windows (and maybe Mac?) for SD cards <=32GB.

lifehackerhansol:
I've also added relevant instructions for SD cards 2GB or less.

lifehackerhansol:
PR is rebased to the latest commit in upstream.

PR #1895, opened 2021-07-23
region-changing: remove link to NNID region change guide

ihaveamac:
It's way too complicated and not actually useful for most people.


fox8091:
Seconding this, and additionally, it's not even up to date with anything post-9.2

lilyuwuu (approved):
LGTM, but the original text is a bit incorrect ("it" is used in reference to both the NNID _and_ the system which could be ambiguous, and "NNID's" shouldn't have the apostrophe)

PR #1897, opened 2021-07-27
Add note on updating system and CFW

lifehackerhansol:
**Description**

This PR:
- adds a note on the latest system version and whether it is safe to update
- renamed Restoring CFW to Restoring/Updating CFW

There are a lot more that could be added, such as it becoming a landing page in case some people have A9LH for some reason or something like that. A tool for checking b9s version would be nice. As this is more preliminary or anything, I would rather this page go through some reviewing before merge.

If you have suggestions ping me on Discord, comment here, or just directly commit.

### DIFF
--- a/_pages/en_US/checking-for-cfw.txt
+++ b/_pages/en_US/checking-for-cfw.txt
@@ -6,6 +6,10 @@ title: "Checking for CFW"
 
 ### Required Reading
 
+The latest system software version is currently 11.15.0-47. After you have completed all the steps in the following pages, it is safe to update.
+Last updated: July 27, 2021
+{: .notice--warning}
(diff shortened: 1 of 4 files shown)


lifehackerhansol:
With suggestion from Discord, I've reworded the homepage notice a little bit. It now doesn't mention safe or unsafe in the homepage at all, but rather in Checking for CFW, which was only changed to add the "safe to update" notice.

I also fixed the link there because apparently that also links to Restoring CFW.

lifehackerhansol:
Checking for CFW system version notice is dropped. Otherwise the same.

lilyuwuu (approved):

PR #1904, opened 2021-08-05
installing-b9s-*: Add note that Luma3DS is booting upon installation of B9S

### LINKED ISSUE #1901: More info about luma3ds
It seems like some users on the discord server asks "does Luma3DS even boot? My home menu looks normal." So I think it would be a good idea to put a mini-disclaimer under the "you will now boot luma3ds by default." which should state something like 
"Disclaimer: luma3ds looks like the normal home menu. To make sure you have luma3DS, follow the next step, or open Rosalina menu."

lifehackerhansol:
**Description**

Fixes #1901. 
Not exactly done with the PR yet, though, only added that Luma3DS is booting after this page.

### DIFF
--- a/_pages/en_US/installing-boot9strap-(browser).txt
+++ b/_pages/en_US/installing-boot9strap-(browser).txt
@@ -53,6 +53,7 @@ new-browserhax-xl/old-browserhax-xl (when combined with universal-otherapp) is c
   + **"Show NAND or user string in System Settings"**
 1. Press (Start) to save and reboot
 
+At this point, your console will boot the Luma3DS CFW by default as long as the SD card is inserted.
 ___
 
 ### Continue to [Finalizing Setup](finalizing-setup)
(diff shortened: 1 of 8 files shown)


lilyuwuu:
"as long as the SD card is inserted" feels a bit odd because it implies that the 3ds will _not_ boot into luma3ds without the SD card (which is technically the truth at this point since the 3DS won't be able to boot, but will be amended in finalizing).  Maybe add a note like "On the next page, you will copy Luma3DS to internal memory so that you can boot without an SD card".

I'd also drop the "CFW" part stylistically (I get that it's used elsewhere in the guide but it feels a bit weird to refer to luma as "Luma3DS CFW"), but that's just me.

lifehackerhansol:
After some discussion a warning box is added to both bannerbomb3 pages that this process may risk uninstalling cfw accidentally.

All pages are also edited to add the Luma3DS to CTRNAND note as well.

PR #1905, opened 2021-08-17
Update MM to 4.24.0 and update some deps

lifehackerhansol:
This PR:
- Updates to Minimal Mistakes 4.24.0 which is the latest. 
  - While not exactly my intentions, this actually reduced build time from 59.718s to **_12.443s_**, so it's literally four times as fast. This may not translate exactly as described as build times are device-specific, but there is no doubt that there _will_ be a performance benefit regardless of what is building, whether that be GitHub Actions or otherwise.
- Updates dependencies to match the latest Minimal Mistakes.
- Adds / updates dependencies to be compatible with Ruby 3. 
  - This will help anyone who is willing to submit changes to the guide as most people will likely install Ruby 3 and updated things etc etc, without having to edit the Gemfile and commit specific files. 

PR is still WIP, as I am attempting to refactor the Gemfile and gemspecs.


lifehackerhansol:
PR is ready for review. While my local testing did not see anything different, it would be nice if this was tested by a few other people locally before merge.

lifehackerhansol:
I've added a dependabot config that updates both `package.json` and the Gemfile. This should automatically update all bundler stuff without having some guy look at the configs and go "wow this is super out of date".

If this is to be reverted please let me know.

PR #1907, opened 2021-08-27
ntrboot things

lifehackerhansol:
**Description**

Still being updated.


lifehackerhansol:
I've kinda lost focus on what this PR was actually supposed to be at this point, but the existing commits are still useful. 

Merged to HEAD.

PR #1915, opened 2021-09-12
Link directly to boot9strap in the main guide path

radostin04:
This PR changes the links to boot9strap in the 3 methods that users will reasonably be following to link directly to boot9strap-1.3.zip, instead of having the user download and choose it manually from the repository.


lifehackerhansol:
Isn't this done in anticipation of future B9S updates? I mean we all know a future B9S update won't happen but still.

radostin04:
Yes it is, but the hassle it causes isn't worth the tiny chance of us having to change the link in event of a b9s update

lilyuwuu:
If there's a critical B9S update in the future then people will need to follow updating-b9s anyway. We'll just have to update each page.

Anyway: I think that we should use direct links for all of the b9s links on the guide (off the top of my head, we're missing links for hbl-soundhax, flashing-ntrboot, and a9lh-to-b9s - updating-b9s too but it might be more appropriate to keep that as the latest link).

radostin04:
I don't disagree, I was just a bit lazy and wanted to get a PR out quickly lol - i'll commit changes to the rest of the pages (except updating b9s, i think that that should stay the way it is rn) sometime soontm

MechanicalDragon0687:
Missing hardmod and troubleshooting pages. Other than that, looks good.

Issue #1918, opened 2021-09-19
fredtool guide still uses steps from old injector site


MechanicalDragon0687:
**Pages with issue(s)**
https://3ds.hacks.guide/installing-boot9strap-(fredtool).html

**Description of the issue(s)**

section 3 steps 4 and 5 are invalid for the current injector page (no captcha, must verify then build)
step 9 also references `fredtool_output.zip` but the filename defaults to `fredtool.zip`

PR #1919, opened 2021-09-20
fix: correct button label on bannerbomb instructions

SethFalco:
**Description**

### BannerBomb
Minor pull request that corrects one of the instructions.

image

The button is labelled `Build and Download`, not `Go`.

### Fredtool

image

---

The final line deletions are from the `.editorconfig` file.

```ini
insert_final_newline = false
```

### Related
* Closes https://github.com/hacks-guide/Guide_3DS/issues/1918


lifehackerhansol (changes requested):
Please only edit the English page. It is futile to change any other language as it will be overwritten by Crowdin and in rare cases can cause other confusion.

lifehackerhansol:
Looks good to me.

If you don't mind, could you also take some time and incorporate all the other issues explained in this related issue: #1918? If you don't want to, I can make that PR myself. Would be nice to fix it in one go, though.

lifehackerhansol (approved):

PR #1934, opened 2021-11-22
How to get Nintendo 3DS folder if it doesn't exist

radostin04:
Sorta wondering if there's a way to mention "if it doesn't work, your formatting might be wrong" in a way that would be easily understandable to users. Any ideas?

### DIFF
--- a/_pages/en_US/seedminer.txt
+++ b/_pages/en_US/seedminer.txt
@@ -18,6 +18,7 @@ This method uses a powerful graphics card to perform the calculations needed. A 
 
 1. Insert your SD card into your computer
 1. Navigate to the `Nintendo 3DS` folder on your SD card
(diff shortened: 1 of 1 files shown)


lilyuwuu:
I don't think mentioning that is necessary. Most people will come into the guide either with an already used 3DS (where the SD card is obviously working) or with a new Japanese 3DS (which will normally include a compatible card, then they Google it and think that only a 32GB card is compatible because that's what Nintendo's support page says).

Not that people read the homepage, but it already says "You will need a working SD card in your 3DS, as well as the ability to write files to the SD card. The 3DS can read SD cards formatted as MBR/FAT32.", too.

Changes made in this PR LGTM, though I'd probably just remove the period at the end of the bullet point because we don't use one anywhere else.

PR #1952, opened 2022-01-11
Use some better colouring on links in notices, also fix table borders

lifehackerhansol:
**Description**

Links look like ass right now as of the rebase, this makes it more like the old guide we had. 

Am willing to take opinions.


lifehackerhansol:
The table borders have been re-added.

PR #1957, opened 2022-01-20
Finalizing Setup rework

lilyuwuu:
This PR reworks Finalizing Setup to (hopefully) make it easier to follow. The following changes have been made:

- The explanation of "valid arm9 payloads" run as `boot.firm` has been removed, as well as references to how SysNAND CFW bypasses the need for EmuNAND.
- The Move EmuNAND infobox has been moved to right before What You Need, and more explicitly says that it doesn't apply to most people.;
- The 2.1.0 CTRTransfer warning has been removed (I don't think an N3DS will even boot on 2.1 with boot9strap, and anyone in that situation would probably ask us for help anyway)
- "What You Need" has been split into three sections, depending on what the user is downloading (CIA files, 3DSX files, and "other")
    - Files are now in alphabetical order, both in What You Need and in Prep Work
- DSP1 is switched from 3DSX to CIA
- Prep Work has been condensed to include all file types in one step
- ctr-no-timeoffset and DSP1 have been combined into one step
- Explicit "Success" message at the very end
- Replace whatever was in "Information and Notes" before with an infobox containing keycombos, a link to the eiphax tips page, and a link to GodMode9 usage

Should be ready to merge, but any suggestions/tweaks/reversions are welcome.

### DIFF
--- a/_pages/en_US/finalizing-setup.txt
+++ b/_pages/en_US/finalizing-setup.txt
@@ -10,14 +10,14 @@ The file `boot.firm` is what is launched by boot9strap itself after it finishes 
 
 On this page, we will make critical system file backups and install the following homebrew programs:
 
-+  **FBI** *(installs CIA formatted games and applications)*
++  **FBI** *(installs CIA formatted applications)*
 +  **Anemone3DS** *(installs custom themes)*
 +  **Checkpoint** *(backs up and restores save files for 3DS and DS games)*
-+  **Universal-Updater** *(a homebrew app store for downloading homebrew from the 3DS over wifi)*
++  **Universal-Updater** *(a homebrew app store for downloading homebrew from the 3DS over Wi-Fi)*
 +  **GodMode9** *(multipurpose tool which can do NAND and cartridge functions)*
 +  **Homebrew Launcher Loader** *(launches the Homebrew Launcher)*
 +  **DSP1** *(allows homebrew applications to have sound)*
-+  **ctr-no-timeoffset** *(removes the RTC offset so that the home menu and RTC timestamps match)*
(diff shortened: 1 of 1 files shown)


_pages/en_US/finalizing-setup.txt
```diff
@@ -57,12 +61,6 @@ During this process, we also setup programs such as the following:
 
 #### Section II - Updating the System
 
-If, while following a previous version of this guide, you CTRTransfered your *New 3DS* to 2.1.0, you should now [restore your NAND backup](godmode9-usage#restoring-a-nand-backup) before doing this section.
-{: .notice--danger}
```
lifehackerhansol, line 61: Is this okay to remove?

lifehackerhansol:
Looks okay other than that one burning question above.

_pages/en_US/finalizing-setup.txt
lilyuwuu: Resolved by re-word: "If your **New 3DS** was on firmware 2.1.0 before following this guide, you should [restore your NAND backup](godmode9-usage#restoring-a-nand-backup) before continuing."

lifehackerhansol (approved):

radostin04 (commented):
I'm not sure how good the "CIA Files" and "3dsx" files headers are over the old method - I think it might get confusing for some helpees.

Maybe we might want to mention the filetype to the right of the link, even for those that only have one file?

lilyuwuu:
I feel like that would be repetitive. People *should* be able to understand "**CIA Files** (get the `.cia` file)" ?

radostin04:
I mean, remove the "CIA files", and keep just the (get the `.cia` file)

lilyuwuu:
resolved

PR #1968, opened 2022-02-04
Introduction V3 - mandatory version checking

lilyuwuu:
**Changes made in this PR:**

homepage: read intro warning -> info
homepage: remove cfw check notices, as it is now part of get-started
homepage: add homebrew software/ports as popular cfw use
homepage: revamp guide installation to describe everything the guide does
homepage: slight reword on device compatibility
homepage: add sd card error check and ban notice

get started: remove cfw check warning
get started: add mandatory cfw and system version check, using "section" structure
get started: remove sd card error check and ban notice (moved to homepage)
get started: convert "version table" to section
get started: add notes on what inclusivity means, how software decimals work
get started: table rework - switch from/to tables to "X to X" language; change "action" to "what to do"
get started: add header for alternate methods
get started: reword alternate methods note

checking-for-cfw: add header to link to from get-started

End result:
[homepage](https://user-images.githubusercontent.com/27717444/152485766-b414296d-aed4-4fae-bfa0-764184ab67dd.png)
[get started](https://user-images.githubusercontent.com/27717444/152485817-d6819f49-314f-4e1b-9ccf-d19c5c1b59c3.png)

### DIFF
--- a/_pages/en_US/get-started.txt
+++ b/_pages/en_US/get-started.txt
@@ -6,60 +6,63 @@ title: "Get Started"
 
 ### Required Reading
 
-If you have hacked your 3DS in the past, or you purchased your 3DS second-hand, it is possible that your console already has custom firmware. You should [check for CFW](checking-for-cfw) before proceeding.
+Before starting this guide, we will see if custom firmware is already is already installed and check the current system version of your device.
+
+#### Section I - CFW Check
+
+1. Power off your device
+1. Hold the (Select) button
+1. Power on your device while still holding the (Select) button
+1. If you do not see a configuration menu (e.g. "Luma3DS Configuration"), you may proceed to the next section
+
+If you see a configuration menu, STOP - you already have custom firmware! Continue from [here](checking-for-cfw#what-to-do-next).
 {: .notice--warning}
 
-Select the appropriate page for your version from the chart below. Note that the "from" and "to" fields are inclusive.
+#### Section II - System Version Check
 
-Your device version can be found at the bottom right of the top screen of the System Settings.
+1. Open the System Settings application
+1. Your system version will be displayed on the bottom right of the top screen (e.g. "Ver. 11.15.0-47U")
 
-![]({{ "/images/screenshots/system-version.png" | absolute_url }})
-{: .notice--info}
+#### Section III - Select a Method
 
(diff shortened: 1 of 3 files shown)


_pages/en_US/get-started.txt
```diff
@@ -6,60 +6,63 @@ title: "Get Started"
...
+      <td style="text-align: center; font-weight: bold;">1.0.0 to 11.3.0</td>
       <td style="text-align: center; font-weight: bold;"><a href="installing-boot9strap-(soundhax)">Installing boot9strap (Soundhax)</a></td>
     </tr>
   </tbody>
 </table>
 
 ---
+#### Alternate Methods
+
+If possible, you should follow one of the software methods listed above.
```
lifehackerhansol, line 63: What if we bold this? Too many people not even bothering to look at the table, make it more noticeable lol

lifehackerhansol (approved):

lifehackerhansol:
Actually, why don't we just force people to check for CFW, like, in complete? 

Home -> Checking for CFW -> Get Started

lilyuwuu:
IMO:
- It would be an additional mandatory page for people to click through. The way that Checking for CFW is structured allows for it to be linked by itself and adding it to the guide path would detract for that since we'd need to explain why people need to check for CFW at the very beginning.
- The page name "Get Started" implies... well, getting started. So people could just skip Checking for CFW since it doesn't seem like it's logically structured within "Get Started -> (method) -> Finalizing Setup".
- Checking for CFW has a lot of infoboxes referring to old CFW things. People shouldn't need to see that if it doesn't apply to them because it is confusing (hence linking to a header on Checking for CFW rather than offering it directly)

I think it kind of fits to check both CFW _and_ system version anyway? Keeps it on one page and makes Get Started feel more consistent with the rest of the guide.

_pages/en_US/get-started.txt
lilyuwuu: Don't think it matters much, "Alternate Methods" already calls attention to them being... alternate methods... and the text is already on a newline (which is a bit weird to bold on its own). At that point we might as well infobox it, but that doesn't really work with the "otherwise" structure of this text.

I do like this wording better than the current one though (thx burguers)

lifehackerhansol:
Makes sense. Don't have complaints otherwise.

radostin04 (approved):

PR #1980, opened 2022-02-18
add twn seedminer method

lilyuwuu:
bb3-fred-twn: new page
bb3: add redirect for twn consoles
seedminer: add bb3-fred-twn method with clear indication and separation to show it is for twn consoles only
main.js: add bb3-fred-twn route
en_us nav: add bb3-fred-twn page
site-nav: add bb3-fred-twn page

Adding directly to guide rather than linking to gist for ease of translation. Hoping the separation and wording is clear enough that no non-TWN consoles will accidentally follow it.

May still need to be checked for 100% accuracy, but at least one person seems to have followed it without issue so far.

### DIFF
--- a/_data/navigation/en_US.yml
+++ b/_data/navigation/en_US.yml
@@ -83,6 +83,9 @@ sidebar_pages:
     title: Homebrew Launcher (Soundhax)
     url: homebrew-launcher-(soundhax)
   -
+    title: BannerBomb3 + Fredtool (TWN)
+    url: bannerbomb3-fredtool-(twn)
+  -
     title: Installing boot9strap (Hardmod)
     url: installing-boot9strap-(hardmod)
   -
@@ -109,3 +112,4 @@ sidebar_pages:
   -
     title: Finalizing Setup
     url: finalizing-setup
+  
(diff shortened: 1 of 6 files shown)


lifehackerhansol:
I'm still wondering if we can just merge all Fredtool methods to use DLP instead. Does it _have_ to be NDS Wi-Fi? The difficulty is going to be the same either way aside from some clicks?

lilyuwuu:
I mean, it _doesn't_, but DSDLP is generally more useful for people and more important to not have broken in the event people forget to to restore it. Jenkins is a bit less user-friendly than the [conventional](https://3ds.nhnarwhal.com/3dstools/bannerbomb3.php) [injectors](https://3ds.nhnarwhal.com/3dstools/fredtool.php) though, so I'd rather we switch it there rather than switch sites.

I think it might be good to merge BB3 and Fredtool into one page (we're long past the days where we'd need to dump DSiWare from HBL) or re-name the BB3 page entirely to make it less likely to be confused with USM-BB3, but that'd be something to resolve sometime else.

lifehackerhansol (approved):

PR #2017, opened 2022-05-18
Correctness fixes for formatting MicroSD cards under Linux

manojka:
It's possible I missed something, but it looks like the existing instructions would have someone create a filesystem on top of the entire card instead of within the partition. This fixes that.


manojka:
I just added another change that pedantically fixes usage of "mount point".

lifehackerhansol (approved):
Yeah, an oversight from the last time I rewrote this. LGTM

PR #2035, opened 2022-06-05
Guide update for Luma3DS v11.0

lilyuwuu:
This PR:

- B9S-*, AtoB, Updating-B9S: Merges the "Installing boot9strap" and "Configuring Luma3DS" sections when possible, as there is no longer anything necessary to configure in Luma3DS.
    - Because B9STool can (probably) still weirdly fail to install B9S sometimes, pages that use it (B9S-Fredtool and BB3-Fredtool-TWN) now have a "Luma3DS Verification" section that prompts people to check for the existence of the `luma` folder and `config.ini`. The next section has people do things on SD anyway, so it works out pretty well.
- B9S-*, AtoB, Updating-B9S: Rewords/clarifies the check for people who can't t boot because of missing `boot.firm`. (Not necessarily related to Luma 11, but useful for clarification purposes as we currently link to a pretty long troubleshooting page).
- B9S-kartdlphax: Renames "SafeB9SInstaller" section to "Installing boot9strap", in line with the rest of the guide.
- Finalizing: Removes boot9/boot11 backup, since Luma should do it now.
- Finalizing: Prompts people to backup `/luma/backups/`.
- Restoring-CFW: Removes Luma3DS configuration *and* GodMode9 Luma3DS to CTRNAND script instructions.
    - While the Luma3DS to CTRNAND script remains useful for troubleshooting purposes (since it copies GodMode9 to CTRNAND's payloads folder, making GodMode9 accessible when an SD is being fishy), it isn't strictly necessary for people that are restoring or updating CFW (since they should have already done that when following Finalizing Setup). Luma3DS to CTRNAND remains part of Finalizing Setup so people will still reasonably be getting GodMode9 to CTRNAND while we figure out what to do with the script situation.

### DIFF
--- a/_pages/en_US/bannerbomb3-fredtool-(twn).txt
+++ b/_pages/en_US/bannerbomb3-fredtool-(twn).txt
@@ -136,9 +136,9 @@ If you would prefer a visual guide to this section, one is available [here](http
   + The existence of these folders and files confirms that Luma3DS is installed
   + If you do not see the `luma` folder or `config.ini`, [follow this troubleshooting guide](https://github.com/zoogie/b9sTool/blob/master/TROUBLESHOOTING.md)
 
-At this point, your console will boot to Luma3DS by default as long as the SD card is inserted.
+At this point, your console will boot to Luma3DS by default.
   + Luma3DS does not look any different from the normal HOME Menu. If your console has booted into the HOME Menu, it is running custom firmware.
-  + On the next page, you will copy Luma3DS to internal memory so that you can boot without an SD card.
+  + On the next page, you will install useful homebrew applications to complete your setup.
 
 #### Section VII - Restoring DS Download Play
 
--- a/_pages/en_US/bannerbomb3-fredtool-(twn).txt
+++ b/_pages/en_US/bannerbomb3-fredtool-(twn).txt
@@ -142,8 +142,6 @@ At this point, your console will boot to Luma3DS by default as long as the SD ca
 
 #### Section VII - Restoring DS Download Play
 
-1. Power off your device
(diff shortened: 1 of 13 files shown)


_pages/en_US/bannerbomb3-fredtool-(twn).txt
```diff
@@ -125,21 +125,23 @@ If you would prefer a visual guide to this section, one is available [here](http
 1. Press (A), then press START and SELECT at the same time to begin the process
 1. Once completed and the bottom screen says “done.”, exit b9sTool, then power off your device
     + You may have to force power off by holding the power button
-    + If you see the Luma Configuration screen, continue with the guide without powering off
+    + If your device shuts down when you try to power it on, ensure that you have copied `boot.firm` from the Luma3DS `.zip` to the root of your SD card
+    + If you see the Luma Configuration screen, press (Start) to reboot, then continue with the guide
 
-#### Section VI - Configuring Luma3DS
+#### Section VI - Luma3DS Verification
```
radostin04, line 131: is there any situation in which the uesr wouldn't see the luma configuration screen after installing luma, that would neccesitate this section?

lilyuwuu: B9STool used to (in rare cases) silently fail to install B9S, hence the [special troubleshooting for it](https://github.com/zoogie/b9sTool/blob/master/TROUBLESHOOTING.md) (see section B). Not sure if this still happens, and also not sure if B9STool consistent And obviously if B9S wasn't installed then the user wouldn't see the Luma configuration menu/etc.

lifehackerhansol (approved):

PR #2067, opened 2022-08-30
kartdlphax: switch to using slotTool

lifehackerhansol:
# Completely untested.

**Description**
universal-otherapp no longer works on NATIVE_FIRM. Switch to using slotTool -> USM.

In theory, this works. We got it working with PicHaxx, no reason for it to not work in kartdlphax.


lifehackerhansol:
moot for now

PR #2068, opened 2022-08-30
update pichaxx to work on latest and add it as an option on the seedminer page

Gruetzig:
Please edit any mistakes, but its good for people that uninstalled cfw in a bad way and cant use bannerbomb/unsafemode


lifehackerhansol:
We had a discussion internally about this already, and ultimately chose not to add this for now. I personally _really_ don't want to bring back HBL-USM. If one brings up a tool to seamlessly do it via otherapp, maybe.

Plus, if we had to consider this particular issue, then we should also be factoring in the issue of people with broken shoulder buttons that can't even use USM in the first place. A fix is to probably use frogtool which in its own is a mess of instructions.

I suggest holding off on this for now.

lifehackerhansol:
We could potentially add it to the wiki, now that I think about it. Not certain if this is worth even reviving.

Gruetzig:
why are you against bringing back that method? i dont see where its dangerous/bad, and for people with broken shoulder buttons there is bannerbomb3+fredtool

Gruetzig:
but i guess youll make your own better version if you decide to add it, it was a crc idea that i make a pull request, was kinda thinking that this would happen

PR #2087, opened 2022-09-14
all: explain in short detail what Luma3DS configuration even is

lifehackerhansol:
**Description**

See title.

Thank you kind stranger for guide feedback.

### DIFF
--- a/_pages/en_US/a9lh-to-b9s.txt
+++ b/_pages/en_US/a9lh-to-b9s.txt
@@ -62,6 +62,8 @@ For all steps in this section, overwrite any existing files on your SD card.
 1. When prompted, input the key combo given to install boot9strap
 1. Once it is complete, press (A) to reboot your device
 1. Your device should have rebooted into the Luma3DS configuration menu
+  + Luma3DS configuration menu are settings for the Luma3DS custom firmware. Many of these settings may be useful for customization or debugging
+  + For the purpose of this guide, these settings will be left on default settings
   + If you get a black screen, [follow this troubleshooting guide](troubleshooting#boot-related-issues-on-modded-devices)
 1. Press (Start) to save and reboot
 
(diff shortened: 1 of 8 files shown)


lifehackerhansol:
Hold until I test this. I don't remember how many nested lists work on Jekyll.

lifehackerhansol:
The answer is 2.

lifehackerhansol:
Small problem with Fredtool and A2B page: because nested levels cap at 2, it's hard to put an explanation on Luma3DS in a subbullet, since the Luma config reference there is already a subbullet.

Would like to RFC on that.

lilyuwuu:
suggested reword (maybe a bit long/TMI?)

  + The Luma3DS configuration menu contains settings for Luma3DS. Many of these settings may be useful for customization or debugging
  + For the purpose of this guide, these settings will be left on default settings, but you may adjust them if you wish
  + The bottom screen will contain information about what the setting does
  + Additional settings can be found in `/luma/config.ini`, but should generally not be touched unless you know what you're doing

(kinda iffy about the last line because don't want people to randomly disable arm11 exception handlers, but this is also probably information that the guide should tell at some point, so maybe in finalizing setup?)

lifehackerhansol:
Can we just put this janky af section into the wiki (along with an improved explanation of the INI since LumaTeam's own wiki is rather lackluster) and just do a `To learn more, visit this page`

lilyuwuu:
The wiki is currently kinda outdated unfortunately (refers to stuff only contained within config.ini)

lifehackerhansol:
> The wiki is currently kinda outdated unfortunately (refers to stuff only contained within config.ini)

which is why we should make use of _our_ wiki to rectify that. It's much better than having to explain what each options are, and we can just do

- The Luma3DS configuration menu contains settings for Luma3DS. Many of these settings may be useful for customization or debugging
- For the purpose of this guide, these settings will be left on default settings, but you may adjust them if you wish
- To learn more about these settings, refer to [this page](somewhereintheworld)

lilyuwuu:
that works

lifehackerhansol:
For now not merging the wiki link.

PR #2091, opened 2022-09-16
bring back kartdlphax

Gruetzig:
kinda tested, last 2 sections untested, should work though

i hope this is not already internally denied

PR #2095, opened 2022-09-20
Add kartdlphax-usm

PabloMK7:
**Description**

This pull request brings back kartdlphax using USM. I have decided not to overwrite the exiting kartdlphax page as it is still a valid method for consoles on firmware lower than 11.16.

### DIFF
--- a/_pages/en_US/get-started.txt
+++ b/_pages/en_US/get-started.txt
@@ -63,5 +63,6 @@ If possible, you should follow one of the software methods listed above.
 
 Otherwise, methods that work on all versions are available, but require additional hardware:
 
+1. [Installing boot9strap (kartdlphax-usm)](installing-boot9strap-(kartdlphax-usm)) - requires a 3DS with custom firmware and a copy of Mario Kart 7
 1. [ntrboot](ntrboot) - requires compatible DS flashcart
 1. [Installing boot9strap (Hardmod)](installing-boot9strap-(hardmod)) - requires soldering
--- a/_pages/en_US/installing-boot9strap-(kartdlphax-usm).txt
+++ b/_pages/en_US/installing-boot9strap-(kartdlphax-usm).txt
@@ -1,0 +1,141 @@
+---
+title: "Installing boot9strap (kartdlphax-usm)"
+---
+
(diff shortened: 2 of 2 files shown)


PabloMK7:
I now just realised pull request #2091 exists ;_;. I have reviewed it and I think there are some sections there that could be improved. Please let me know how to proceed.

lifehackerhansol:
IMO, we should only have one kartdlphax page, so I'm more in favour of the other PR.

You can suggest changes there.

PabloMK7:
The problem with the other PR is that it completely overwrites the original kartdlhax page. Suggesting changes there would be totally overwriting the existing changes.

PabloMK7:
The problem is not with the diff. The other PR replaces the existing kartdlphax instructions with the new ones, while this PR creates a different page. The other PR contributor would need to restore the existing page and add the same txt file in this PR

lifehackerhansol:
I'm not exactly certain if we want another kartdlphax page. 
@lilyuwuu what's your opinion?

lilyuwuu:
As it stands, if we're only going to use USM with kartdlphax, then it's probably better to keep it as "Installing boot9strap (kartdlphax)". On the other hand, based on guide statistics, about 20% of people are doing BB3-Fredtool, presumably due to broken shoulder buttons, so we could presume that anyone doing kartdlphax would end up in the same situation.

lifehackerhansol:
I think the goal of this PR is less about having an archived version and more having the 11.15 version publicly facing so people on 11.15 can just go straight to universal-otherapp instead.

I'm of the argument that it's not necessary to keep both versions of kartdlphax, particularly since kartdlphax itself is quite literally _never_ used. Seedminer is still an invincible method and kartdlphax will always be a "if Seedminer doesn't work do this" method. In fact, the only time I've ever recommended anyone to follow it was people with a banned console or a launch model O3DS which is fundamentally incompatible with Seedminer. This would mean that a very large chunk of users updated their console to follow Seedminer and failed. Hence I don't see many people on 11.15 and lower following it.

PabloMK7:
Closed as suggested

PR #2099, opened 2022-09-24
pichaxx redirect applied to usm but not bb3

Gruetzig:
When you have broken shoulder buttons you go to bb3, and then you also cant use that method because of broken dsiware management, rare but possible and then this addition is important, unless you move the button warning back down on the usm page

### DIFF
--- a/_pages/en_US/troubleshooting.txt
+++ b/_pages/en_US/troubleshooting.txt
@@ -148,7 +148,9 @@ There may be an issue with your `F00D43D5.bin` file (it may be corrupted or inte
 
 ### DSiWare Management menu crashes without purple screen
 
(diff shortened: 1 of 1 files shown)


lifehackerhansol:
I'm not certain what this accomplishes.

Fredtool redirect is done for broken shoulder buttons, and pichaxx redirect is done for broken DSiWare management, yes? But the main is still USM.

Only reason one would be sent to fredtool is because buttons don't work in USM. Obviously, pichaxx also will have the same issue. 

So if some poor soul followed BannerBomb3 first, they are watching a video guide. Fredtool isn't linked as a main guide path, unless your device is TWN it is always a hidden alternate route only accessible if SAFE_MODE does not work.

That rare chance someone has working buttons but following BB3 first is someone not reading the guide.

Gruetzig:
i actually forgot that pichaxx needs shoulder buttons

PR #2103, opened 2022-10-01
super epic omega very cool explanations and pics

lilyuwuu:
Lots of changes here!

- Switch from "Required Reading" to "Technical Details" as a collapsible, with more technical details (or links to the actual page about the exploit for anyone who wants it.
- Add a "compatibility notes" section for information on what is necessary for the exploit (region, shoulder buttons, etc)
- Add an explanation of what the user is actually doing below every section.
- Add some images - particularly: the "pasting id0 somewhere" thing, friend registration, images of the location of the Nintendo DSiWare folder, DSiWare Management, USM menu
- Slightly reorganize photos on repo so that each page has its own folder. (but only kinda sorta because this got annoying and we still have more photos to add at some point)


lifehackerhansol:
Removed the text-expand includes as it is not needed.

Technical explanation has been isolated into a info-notice box. Purely visual change, the JS hack still works. Much tidier this way, though.

lifehackerhansol (approved):

Issue #2108, opened 2022-10-07
no 11.16 firms on hardmod page


foxsouns:
**Pages with issue(s)**
https://3ds.hacks.guide/installing-boot9strap-(hardmod)

**Description of the issue(s)**
Page needs to be updated to display a new firm or update text. (Need find out if 11.16 uses a new kernel version (not likely) and update accordingly.)

lifehackerhansol:
I believe there is a new nfirm. We haven't made a new torrent for it yet.

Perhaps I'll throw a warning on the page for now until I get my hands on it.

foxsouns:
@lifehackerhansol Oh, I think could make the warning! I'll make the PR.

PR #2109, opened 2022-10-07
add warning against installing old firm files

foxsouns:
**Description**

Hardmod firms for the `installing b9s - hardmod` page are not updated: this includes a warning to disuade readers from installing incorrect firmware until #2108 gets a torrent created.


foxsouns:
oh no, i was too late! alas, closing PR.

lifehackerhansol:
ninja'd

PR #2120, opened 2022-10-14
clarify hardware button check

Gruetzig:
"I am in the hardware safe mode testing thing rn it prompted an update and after I pressed cancel it didn't go to the home menu it just shut off is this fine?"
"in this section of the guide my 3ds does go into the system update menu but doesn't go back to the main menu after I press the cancel button should I still do the safe mode method or the other one ?"
should fix these questions


lifehackerhansol:
How about we just combine it into one subbullet
```
1. If prompted to update, press Cancel
  + Your device will power off
  + If the device boots to the HOME Menu, just power off your device
```

Gruetzig:
good idea ;)
(had the idea too but something stopped me)

PR #2128, opened 2022-10-29
It's actually not really easy to hack your 3DS. Not everyone can do it.

Gruetzig:
yes(this sentence should really get removed)

### DIFF
--- a/_pages/en_US/home.txt
+++ b/_pages/en_US/home.txt
@@ -46,7 +46,6 @@ This guide will:
 
 ## What do I need to know before starting?
 
-* It's actually really easy to hack your 3DS. Anyone can do it.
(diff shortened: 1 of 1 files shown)


lilyuwuu:
the current sentence is a reference to a meme and should either be kept as is or removed imo. Changing it to "read the guide lol" is a bit too passive aggressive

adrot:
Agreed. The "fix" feels self-defeating. Taken at face value, if the user cannot read, then the Guide serves no purpose to them. Plus, this makes it look like we are looking down on people if they are having issues.

lifehackerhansol:
You know what, fuck it. We've had our fun with the meme. I'm just going to remove it.

PR #2130, opened 2022-10-30
clarify cfw check

Gruetzig:
(sent picture of the cfw check)
smol💀💀💀💀 — Today at 7:17 PM
did you do it or not
nah — Today at 7:17 PM
no
smol💀💀💀💀 — Today at 7:17 PM
why not
its the first section of the guide
nah — Today at 7:17 PM
i didnt think i had to bc i never got to actually downloading anything

### DIFF
--- a/_pages/en_US/get-started.txt
+++ b/_pages/en_US/get-started.txt
@@ -4,7 +4,7 @@ title: "Get Started"
 
 {% include toc title="Table of Contents" %}
 
-Before starting this guide, we will see if custom firmware is already installed and check the current system version of your device.
+Before starting this guide, we will see if custom firmware is already installed and check the current system version of your device. Do not skip this, no matter of what happened to your device or the SD card in the past. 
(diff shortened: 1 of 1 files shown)


lilyuwuu:
i'm kinda iffy on this - it seems kind of unnecessary and anyone who is purposely skipping it won't necessarily suddenly not skip it just because they potentially see a line

Gruetzig:
maybe it is unnecessary but its not harmful and even if it doesnt help 99,9% of all people it still helps 0,1%.

lilyuwuu:
there is potential harm in that it's kind of unnecessarily scary the way it is currently worded

Gruetzig:
where is it scary

lilyuwuu:
"do not skip this" implies that something bad will happen to the console if it is skipped, which is rarely the case

Gruetzig:
fine

PR #2137, opened 2022-11-09
create troubleshooting for jp n3ds error 032-0420

Gruetzig:
thanks to luigoalma for mentioning this fix

### DIFF
--- a/_pages/en_US/troubleshooting.txt
+++ b/_pages/en_US/troubleshooting.txt
@@ -132,6 +132,10 @@ Browser based exploits (such as this one) are often unstable and crash frequentl
 
 The SSLoth proxy was incorrectly configured. Re-do the SSLoth section on the page.
 
+### Error 032-0420 when trying to open the browser
+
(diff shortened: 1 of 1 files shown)


lifehackerhansol:
Thoughts on just making this a step in the guide?

Gruetzig:
do you mean something like:
`If you receive error 032-0420, follow [this troubleshooting guide](link to troubleshooting)`?

Gruetzig:
or add this long troubleshooting to the page directly

Gruetzig:
I dont see an existing troubeshooting that can be combined with the new one

lifehackerhansol:
> I dont see an existing troubeshooting that can be combined with the new one it doesnt crash, no fancy flashing, no update prompt

I'd argue an error code is a crash, but sure. (what do we define as a "crash" anyway?)

I guess we will consider the other error codes later. I think this is fine.

lifehackerhansol:
I rewrote this troubleshooting a bit to clean up. Does it look alright?

Gruetzig:
a bit, lol, yeah looks good

PR #2139, opened 2022-11-11
yet another try to make ssloth better

Gruetzig:
yattmsb


lifehackerhansol:
This troubleshooting, did it actually happen somewhere?

Gruetzig:
yes https://discord.com/channels/196618637950451712/196635695958196224/1040551938799972373

lifehackerhansol:
I wonder which error this is, an error on the HOME menu or a full on system exception... That helpee's description is _really_ not detailed at all.

Well I found some ancient crc documentation by @lilyuwuu so maybe it's time to add most of those.

PR #2156, opened 2022-12-05
Dropdowns: part 2

lifehackerhansol:
**Description**

Re-create PR because it was accidentally merged, then reverted.


lifehackerhansol:
merged

PR #2159, opened 2022-12-14
Remove unused pages 3: crowdin sucks and we have to clean up after them

Gruetzig:
delete 3 files from all languages, homebrew-launcher soundhax, the thing that comes after that and boot9strap pichaxx


lifehackerhansol:
+1

PR #2161, opened 2022-12-18
remove troubleshooting for missing essentials.exefs

Gruetzig:
1. its called essential.exefs
2. steps are part of finalizing
thank you


lifehackerhansol:
the troubleshooting was always there, the step in finalizing was added later (by me) and we just kinda forgot troubleshooting exists ~~because no one reads it~~

PR #2148, opened 2022-11-23
basic-information.txt

lifehackerhansol:
take 2

images not found

PR #2166, opened 2022-12-30
prevent seedminer things from going wrong :P

Gruetzig:
-add flashcart sd notice, seems to be an issue, example from NH https://discord.com/channels/196618637950451712/247557068490276864/1045096506807570432 
-add steps for generating a Nintendo 3DS folder or the actual ID0 if it hasnt been generated yet this is never done and some people use the wrong ID0 apparently

### DIFF
--- a/_pages/en_US/seedminer.txt
+++ b/_pages/en_US/seedminer.txt
@@ -21,6 +21,10 @@ This method uses a powerful graphics card to perform the calculations needed. A 
 
 In this section, you will get the necessary details from your 3DS that are required to figure out your device's encryption key.
 
+1. Insert your SD card into your device if it's not inserted already
+  + Use the SD card slot from the console, NOT from your flashcart if you have one
+1. Turn the device on if it's not on already
+1. Turn the device off
 1. Insert your SD card into your computer
 1. Navigate to the `Nintendo 3DS` folder on your SD card
   + If you don't have a `Nintendo 3DS` folder, put your SD card into your 3DS and power it on so that the folder can be created


lilyuwuu:
Not personally a big fan of the changes here.

- Most people don't have flashcarts. It isn't reasonable for us to potentially confuse people (by talking about flashcarts, which most people don't have) by adding lines to account for uncommon edge cases.
- People who don't have a Nintendo 3DS folder are already covered by the "If you don't have a Nintendo 3DS folder" bullet point. Adding "power on then power off" instructions for all users is kinda clunky and doesn't account for people who don't have an SD card or whose SD card isn't properly inserted or something (hopefully to be fixed if/when #2148 is merged).

@lifehackerhansol thoughts?

Gruetzig:
> Most people don't have flashcarts. It isn't reasonable for us to potentially confuse people (by talking about flashcarts, which most people don't have) by adding lines to account for uncommon edge cases.

That's true, did think of that but wanted to hear your thoughts.
> People who don't have a Nintendo 3DS folder are already covered by the "If you don't have a Nintendo 3DS folder" bullet point. 

Yes, but I was more focusing on generating the correct ID0, because I often see people in the assistance channels saying
something like "When I did Seedminer I only had one ID0, but then a second one appeared and I ignored it", shouldnt happen unless the 3DS hasn't been booted up with SD yet. Then they place the Nintendo DSiWare folder in wrong id0 and generate a wrong movable and f00d43d5.bin file which can be prevented with these steps. Only having steps for missing Nintendo 3DS folder is not enough in my opinion.

lifehackerhansol:
Man I wonder where all these people get their extraneous ID0s...

There isn't really a clear-cut solution for this. Yes a power off and power on solves this. But that would just make multiple power on and off for people who end up with two ID0s by doing that.

Why can't we just force making a backup of Nintendo 3DS right at the start? That way it's somewhat a clean slate? I don't really like this idea either but it's better than having to reboot like 5 times just to figure out a 32-char number

MechanicalDragon0687:
if they get a valid movable.sed, then the id0 is for their system meaning it went through a system transfer or format and was not booted to home menu after. This is a niche case and i dont think we need extra steps to handle this

Gruetzig:
I guess my ideas are all bad :(

PR #2167, opened 2023-01-04
Create Installing-hShop.txt

Minionguyjpro:
Added the first method for installing hShop/3hs,

**Description**

### DIFF
--- a/_pages/en_US/Installing-hShop.txt
+++ b/_pages/en_US/Installing-hShop.txt
@@ -1,0 +1,31 @@
+---
+title: "Installing hShop"
+---
+
+{% include toc title="Table of Contents" %}
+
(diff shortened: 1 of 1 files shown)


lilyuwuu:
This guide will not add any instructions or references to piracy tools.

Minionguyjpro:
> This guide will not add any instructions or references to piracy tools.

May I ask what you mean? Do you mean, it is not relevant, as it is just an application? I added another commit with a warning and a description, which says it comes in handy as the eShop is going to end in March 2023.

ihaveamac:
It actively enables downloading of games and software that the user didn't purchase. It's a piracy tool, something the guide isn't going to support. Yes, even when the eShop goes down, it's still considered piracy.

PR #2178, opened 2023-01-21
make basic-information a mandatory page

lilyuwuu:
- Re-add Get Started button at the top of the homepage
- Rename "Basic Information" to "Key Information"
- Remove interim notice from Get Started

TODO/To discuss:
~~- Remove skip button from top of Key Information page? (pros: people who know what they're doing save some time. cons: people who don't know what they're doing might click the button by mistake, and it adds a bit of bulk to the page.)~~ Keeping for now
~~- Consistency of using "device" or "console" throughout the guide (let's probably use device for now and switch all guide pages to "console" later if we decide to do that)~~ nothing to resolve
~~- Does this break the left sidebar? Is the Key Information page added to the sidebar so that people with JS/cookies(?) on can use it?~~ Preliminary fix
~~- Should Get Started be renamed? The "Get Started" button page on the homepage links to a page that is not "Get Started" and then there is also a page that exists called "Get Started"~~ nah


lifehackerhansol (approved):
i'm gonna assume the js is working, because I don't feel like testing

image

PR #2181, opened 2023-01-23
add parantheses around re

Gruetzig:
people may think they have to add the bot if they never did it before(neat case but no downsides right)


lifehackerhansol (approved):

lilyuwuu:
Actually, there is a downside:
image

Gruetzig:
thank you lily

PR #2205, opened 2023-03-04
Add includes for ctrtransfer steps

lifehackerhansol:
Pages using the CTRTransfer sections:
- ctrtransfer
- region-changing

Pages using the NAND backup section:
- godmode9-usage
- region-changing

Pages using GM9 SD format section:
- godmode9-usage
- move-emunand

Pages using the HBL launch via DLP section:
- finalizing-setup
- installing-boot9strap-(hbl-usm)

### DIFF
--- a/_pages/en_US/ctrtransfer.txt
+++ b/_pages/en_US/ctrtransfer.txt
@@ -41,86 +41,23 @@ To use the [magnet](https://wikipedia.org/wiki/Magnet_URI_scheme) links on this 
 
 #### Section I - Prep Work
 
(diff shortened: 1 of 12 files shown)


lifehackerhansol:
Added a real guide change:
- We aren't setting up an emuNAND anymore, just tell them to pick that explicitly instead of being vague
  - if someone really wants to set it up it's really not that hard anyway, and they also prob have a real reason to
- We aren't going to make use of the eject thing, just have them power off on each section, turning off console seems easier to do than explaining ejecting

PR #2209, opened 2023-03-10
was fun while it lasted

lifehackerhansol:


lilyuwuu (commented):
might also wanna remove the path from _main.js

lifehackerhansol:
> might also wanna remove the path from _main.js

done

lilyuwuu:
might wanna unremove hbl-usm from this pr pending recent releases(?)

lifehackerhansol:
hbl usm left intact for now, but removed from js as it is currently unused

we can add it back there once we have a guide path

PR #2211, opened 2023-03-12
fix some small issues

Gruetzig:
noone is perfect :dagger:


Gruetzig:
why is this a draft whoops

PR #2215, opened 2023-03-14
Fix twn seedminer redirection

danny8376:
Fix twn seedminer redirection.

~~I know that no one care about TWN consoles~~


danny8376:
close this for now, going to also fix some other issue as well.

PR #2217, opened 2023-03-15
hueheuehueheuhe

eip618:
**Description**

### DIFF
--- a/_pages/en_US/bannerbomb3-fredtool-(twn).txt
+++ b/_pages/en_US/bannerbomb3-fredtool-(twn).txt
@@ -53,6 +53,7 @@ These instructions are for Taiwanese consoles ONLY (as indicated by a T at the e
 1. Navigate to `Nintendo 3DS` -> `<ID0>` -> `<ID1>` on your SD card
   + The `<ID0>` will be the same one that you used in [Seedminer](seedminer)
   + The `<ID1>` is a 32 character long folder inside of the `<ID0>`
+  + If you have further difficulty identifying the ID0, ID1 or where to put the `F00D43D5.bin` file, please watch [this video](https://www.youtube.com/watch?v=T67kcO3wvHY) for a more in-depth explanation{: .notice--info}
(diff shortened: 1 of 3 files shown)


lilyuwuu (commented):
If in a notice, the sentence should end with a period. The "please" not strictly necessary either.

As was discussed elsewhere, I'm personally not for the video as it stands now. I don't really see how this is going to be much more helpful than a gif (or video) depicting someone going through the folders. More importantly, the video as-is seems to over-explain some things (i.e. the whole "ID0 and ID1 are symbolic names" thing) and explains things slower than we probably would in assistance channels (e.g. "go into the Nintendo 3DS folder, you see that long folder name, that's your ID0, go into it, you see that, that's your ID1, go into it again, you see the extdata folder, that's where you make the Nintendo DSiWare folder, yes there")

That's just my opinion though, would like others' input

Gruetzig:
I agree with lily here(what a surprise), I understand that you want the ad revenue, but a gif in soemthing like a drop-down is better than a 4 minute video.

eip618:
> If in a notice, the sentence should end with a period. The "please" is not strictly necessary either.
> 
> As was discussed elsewhere, I'm personally not for the video as it stands now. I don't really see how this is going to be much more helpful than a gif (or video) depicting someone going through the folders. More importantly, the video as-is seems to over-explain some things (i.e. the whole "ID0 and ID1 are symbolic names" thing) and explains things slower than we probably would in assistance channels (e.g. "go into the Nintendo 3DS folder, you see that long folder name, that's your ID0, go into it, you see that, that's your ID1, go into it again, you see the extdata folder, that's where you make the Nintendo DSiWare folder, yes there")
> 
> That's just my opinion though, would like others' input

i made the video the way i did because it addressed what i had assessed to be the most common questions seen recently in the channels. it is/was my objective/intent that the more we can educate people, the less issues we will have in the future.

having said that i will spend some time tomorrow making a simpler, more direct gif

lilyuwuu:
for better or for worse we don't need to find the id0 id1 nintendo dsiware f00d43d5.bin anymore

PR #2219, opened 2023-03-16
Add hardmode 11.16 firm

### LINKED ISSUE #2123: hardmod firm for 11.16.0 still missing
**Pages with issue(s)**

https://3ds.hacks.guide/installing-boot9strap-(hardmod)

**Description of the issue(s)**

Page still needs someone to find that hardmod firm. Commit `b209c7e` only added a warning, this is still something that needs to be solved.

danny8376:
Resolve #2123
Hope that i can seed properly before others also start seeding?
rss.xml is... newly generated with these two file only and then merge manually, seems to be correct... but maybe check again?


lifehackerhansol:
The PR itself is fine, as long as the magnet works.

Can you DM me the files, I want to sanity check these before merging.

lilyuwuu:
.r11

rss.xml
```diff
@@ -754,6 +754,156 @@
 				</group>
 			</trackers>
 		</item>
+		<item>
```
lifehackerhansol, line 757: adding this manually is fine, but it's an RSS feed, so we want newly added items at the top of the RSS, not somewhere in the middle.

lifehackerhansol (changes requested):

rss.xml
danny8376: Ah, indeed, fixed

lifehackerhansol:
last thing before this merges

can you please update the zip here: https://github.com/hacks-guide/Guide_3DS/blob/master/assets/torrents.zip

just add the two new .torrent files and rezip it

lifehackerhansol (approved):

lifehackerhansol:
thankfully, next time we don't have to manually add it, I made the script properly sort it now

danny8376:
Welp, found that typo hardmode all the time, at least it's not inside commit message.

PR #2222, opened 2023-03-18
Make "Console Issue" template redirect to NH

lifehackerhansol:
**Description**

Untested, awaiting feedback.

Instead of "Console Issue" and it creating an issue, it'll just direct you to NH.


lifehackerhansol:
It'll look something like this (select new issue here) https://github.com/lifehackerhansol/Guide_3DS/issues

NightScript370 (approved):

PR #2228, opened 2023-03-22
Add soundhax-sp-twn and fix troubleshooting and navigation

danny8376:
Add soundhax-sp-twn
Fix navigation list for kor
Add/modify soundhax-sp related things in troubleshooting


danny8376:
> maybe it tries abghax dunno
That's probably the case... as it's decided by kernel version, which likely to be updated by cart update...
Actually, that make think... are those frankenfirmware really exist?
Probably time to check what's the latest cart update possible on this console.

Still need to handle nver <= 7 for soundhax, but just within troubleshooting should be just enough.

lifehackerhansol:
If it's unconfirmed, then don't add it.

danny8376:
well, found that soundhax-sp is interesting (and annoying), time to nuke this ~~and probably all soundhax-sp thing~~

PR #2231, opened 2023-03-25
soundhax-sp fix

danny8376:
For reference, https://gist.github.com/danny8376/524394a41202c3617e6a1ef6c181fa91

* make soundhax-sp always use post-5.0 file since we only bring people to soundhax-sp if soundhax won't work, aka version above 11.3, which pre-5.0 file will always crash
* remove soundhax-sp for version below 6.0.0-11, as this is the least version with compatible agb_firm (post-11.3 always use agbhax) and cart updates seems never offer update for agb_firm
* remove soundhax-sp-kor because of the above two reason, as it's kind of redundant now
* remove soundhax-sp redirection for twn as the agb_firm is incompatible with post-11.3 hax
* make troubleshooting of ssloth-browser use soundhax-sp instead of standalone instructions
* add discord link for troubleshooting of  "could not play" aside from seedminer for old 3ds, as those console might be unminable, especially that they have such early cart-updated system.

Might need more test? Although I'm already pretty confident of my conclusion currently.


_pages/en_US/troubleshooting.txt
```diff
@@ -310,6 +310,8 @@ There is an issue with your `otherapp.bin` file (it is missing, misplaced, or co
 <summary><u>"Could not play"</u></summary>
 
 You have the wrong Soundhax file for your device and region, or your device is incompatible with Soundhax. In the latter case, you should update your device to the latest version and follow [Seedminer](seedminer).
+
+If your device is original 3DS or the data is transfered from original 3DS, join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp) for assistance.
```
lifehackerhansol, line 314: what's the rationale behind this one?

danny8376: * add discord link for troubleshooting of "could not play" aside from seedminer for old 3ds, as those console might be unminable, especially that they have such early cart-updated system.
Well, currently, 3ds at launch (original 3ds only ofc) is known to be unminable.
From my test, unless you have very early sound app (v0 specifically), it's not likely to show unplayable, it'll just crash.
(Well, unless they're trying to do soundhax on 11.4.0, sound v4096, which they shouldn't reach soundhax page at first place)
And if you have v0 sound app, it's really likely to be launch 3ds imo. (3ds that have 2.2.0 from factory is likely to have sound v1027, which is not really far from the release date)
From my test again, special build of soundhax should be feasible on those console. It's better for them to try this first before they update and try seedminer (and find that it's unminable and then buy flashcart)

Oh, but after review it again myself, should be original 3DS only, those transfered console should be just soundhax compatible, unless the console itself is also original 3DS and having same issue.

lifehackerhansol:
Superseded by: https://github.com/hacks-guide/Guide_3DS/pull/2237

Sorry, we're just not doing these anymore. Let's stick to doing things normally, and just using another exploit in cases where these kinds of hacks are really necessary.

PR #2233, opened 2023-03-26
Revert the version selector

lifehackerhansol:
**Description**

This seems to be having the opposite effect of the original intent of the selector, which was to make it easier for users by just entering the version and they get magically sent to the correct place.

After some discussion with @Gruetzig this doesn't seem to be worth maintaining, especially after this script started being abused to add all the different soundhax methods, which to be quite honest one is better off doing safecerthax or seedminer by that point.

This reverts the selector entirely, while also leaving some of it behind to keep it working for translations.


Gruetzig (approved):
may want to readd a way to get to Bannerbomb3 fredtool twn

lifehackerhansol:
> may want to readd a way to get to Bannerbomb3 fredtool twn

This should be discussed in another PR. We removed it because far too many people were following the twn stuff without actually being twn. If needed, should be an inline troubleshooting or something (which admittedly makes it harder for actual TWN users but eh.)

lilyuwuu (approved):

eip618:
unpopular opinion i know but i actually like the selector, it should help to avoid the 11.16==11.6 sort of thing we keep seeing

lifehackerhansol:
Closing for now. It's still useful for things like TWN etc.

PR #2235, opened 2023-03-26
Include info on multi console

SirGamers:
kartdlphax can do multi console but the success rate is low each time you add a console. Added info on that


lifehackerhansol:
Is... is this really necessary?

PR #2237, opened 2023-03-29
Remove Soundhax derivatives

lifehackerhansol:
This is unnecessary and causes confusion for both the end user and the helper, when one can just do another exploit if in such a situation where these are necessary.

Exploits to be considered in such a case:
- O3DS: safecerthax
- N3DS: update and do seedminer/browserhax


lilyuwuu (approved):
Lgtm

PR #2239, opened 2023-03-29
region-changing: data yeet, reword

lilyuwuu:
- Reword several introductory parts, including more specific information (i.e. when a region change should be performed, what exactly a region change 'breaks')
- Switch from system format notice to mandatory ~~system save data~~ cfgsave removal (this is to prevent issues on o3ds while also being less destructive in such cases, hopefully - also, since these are consoles that are likely to be imported, it is unlikely for a significant amount of system data to be important to the user)
- Add success notice on bottom (cosmetic)
- Merge sections VI and VII (section VI also takes place in godmode9, making VII redundant)

~~Pending discovery of root cause of extended memory mode and exiting friends list / game notes breaking. If the cause is found and easy to implement (i.e. deleting a single file instead of the folder), we can switch to doing that instead~~

Cause found to be cfgsave, guide is now less destructive

### DIFF
--- a/_pages/en_US/region-changing.txt
+++ b/_pages/en_US/region-changing.txt
@@ -6,11 +6,15 @@ title: "Region Changing"
 
 ### Required Reading
 
-This is an add-on section for region changing your CFW SysNAND. This is done by installing the 11.15.0 CTRTransfer image for the region you want to switch to. After installing the image, you can update your device normally to the latest firmware (11.16.0).
+This is an add-on section for region changing your device. This is done by installing the 11.15.0 CTRTransfer image for the region you want to switch to. After installing the image, you can update your device normally to the latest firmware (11.16.0).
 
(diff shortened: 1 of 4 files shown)


lifehackerhansol:
This needs some changes in ctr include, will fix that in a bit

lifehackerhansol:
Ok I'm dumb, I retract previous statement

_pages/en_US/region-changing.txt
```diff
@@ -72,15 +72,29 @@ To use the [magnet](https://wikipedia.org/wiki/Magnet_URI_scheme) links on this
...
+1. Press (Right Shoulder) + (A) to bring up the folder options
+1. Select "Copy to 0:/gm9/out"
+1. Press (A) to continue
+1. While still highlighting `00010017`, press (X) to delete it
+1. Press (A) to confirm
+1. Press (A) to unlock SysNAND (lvl2) writing, then input the key combo given
+1. Once the file has been deleted, press (A) to continue
+1. Press (A) to relock write permissions if prompted
+1. Press (Start) to reboot your device
+1. Complete the 3DS initial setup as normal
```
lifehackerhansol, line 95: Could use some more explaining for end user, otherwise seems to be ok overall

lifehackerhansol: initial setup

guess it really doesn't matter either way, just thought it would be nice to add at least a small note that it boots into first time setup and this is normal and you didn't lose data
small reassurance goes a long way they say

lifehackerhansol (approved):

PR #2246, opened 2023-04-08
build main.min.js on the fly, add README

lifehackerhansol:
The README shows documentation on how to test the site locally and where to go for translations.

This change will build main.min.js from GitHub Actions, instead of having to manually change it locally.

Breaking change for local testers: `main.min.js` is dropped. Build it manually.


lifehackerhansol:
I'm going to opt to not deviate from upstream for now.

PR #2249, opened 2023-04-14
finalize: switch to auto-script

lilyuwuu:
This PR switches much of Finalizing Setup from user input to a GodMode9 script, reducing the page from 71 to 17 steps.

Benefits:
- Files are packaged in an AIO, no longer requiring users download specific files from GitHub and then place them in specific locations
- The guide has a script that can be updated on its own schedule, rather than waiting for a GM9Megascript update + GodMode9 release (should that ever be necessary)
- Less capacity for user error: If the AIO pack is not copied properly, GodMode9 is unlikely to even open
- Broken shoulder buttons resulting in an inaccessible Rosalina menu are handled by the script
- Setup is significantly simplified, requiring just 20-ish button presses between the console being off and the end of the script

Shortcomings:
- **The script is only available in English**. It could be adapted to other languages, but this would require maintaining separate scripts, which would become a maintenance nightmare. Errors caught by the script are clearly numbered and their troubleshooting re-iterated on the Troubleshooting page to hopefully mitigate this being an issue.
~~- Not every possible error is caught. One notable possible error is someone NOT going to the "reset data management" prompt after the dummy file is created (Error # 05). The script will then appear to continue, but fail at the homebrew installation step (since the title database is not real). This is currently accounted for in troubleshooting, but could be dealt with more elegantly (either by supplying a blank title database or by somehow adding a flag that would be removed once those files are not their dummy sizes).~~ Error is now caught.
~~-  **At the moment, RTC nullification and DSP setup have been removed**.~~ Re-added.
    ~~- DSP setup can be easily re-added as a separate section, but it would be ideal if its dumping could also be automated. I removed it thinking that it isn't absolutely necessary and could be added as a "things to do after" or troubleshooting sorta thing, but it would also make sense to just do in the guide.~~
    ~~- RTC offset nullification cannot logically be re-added very easily, since GodMode9 is now launched before RTC can be nullified. However, since the original justification was for GodMode9 to display dates properly and GodMode9 appears to display dates properly anyway when RTC is set properly *and* it isn't even recommended due to causing ""Lumabug"", I think we can live with removing it.~~
 
Script has been tested on a US N3DS. ~~Marking as draft as we decide whether we should even do this and as this is tested on other regions and models.~~ Ready for review.


Gruetzig:
sus bro, now people wont know how their system works at all and just be thrown into it, not even knownhoe to use fbi to install questionable apps

ihaveamac:
The assets should be packed into an archive that GodMode9 can mount, like a RomFS or tar(?). This makes it easier for users as they only have to copy two files to the SD card, removing the need to merge directories and misplace some files.  would also allow for including additional support files for the script, like PNGs to display.

lilyuwuu:
Images can be displayed without using an archive: `set PREVIEW_MODE "0:/finalize/test.png"` for example.

I'm not sure which two files you mean for the archive thing. To get into GodMode9, we'd either need to make GodMode9 our boot.firm (which would mean we'd be bootlooping into the script should the script be unable to handle something, which leaves people in a seemingly-bricked state which we should avoid) or copy it to `/luma/payloads/` (which introduces user error, such as getting the wrong GodMode9 file or misnaming "payloads").

Most users coming from a vanilla setup wouldn't have much to merge to begin with (they wouldn't have gm9, "finalize" never would have existed, making the only thing to merge `luma`, which would introduces error regardless since our two options are to merge the folders or to have people create the payloads folder).

So, I can't think of much benefit for packaging stuff into an archive for GodMode9 to deal with.

ihaveamac:
> Images can be displayed without using an archive: `set PREVIEW_MODE "0:/finalize/test.png"` for example.

To clarify I don't mean that an archive is required, just that it would make it something that the user doesn't need to manage or clean up later. Or something that our script has to deal with if someone leaves older files on the SD card, since all the expected stuff would be in an archive.

> I'm not sure which two files you mean for the archive thing. To get into GodMode9, we'd either need to make GodMode9 our boot.firm (which would mean we'd be bootlooping into the script should the script be unable to handle something, which leaves people in a seemingly-bricked state which we should avoid) or copy it to `/luma/payloads/` (which introduces user error, such as getting the wrong GodMode9 file or misnaming "payloads").

Then the system I was thinking of is actually three files:
`/luma/payloads/GodMode9.firm`
`/gm9/scripts/finalize.gm9`
`/finalize.romfs` (theoretical)

Although we could have an autorun script put as boot.firm that's a whole discussion for another time.

> Most users coming from a vanilla setup wouldn't have much to merge to begin with (they wouldn't have gm9, "finalize" never would have existed, making the only thing to merge `luma`, which would introduces error regardless since our two options are to merge the folders or to have people create the payloads folder).

I still think that this would reduce user error, even on an empty SD card, since users are still prone to misplacing stuff and reducing the amount of things that could be misplaced would help everyone out.

In addition, not everyone following Finalizing Setup is actually coming from a stock console, some people use it to upgrade their setup (like from b9s <=1.1 or a9lh) or do so after a System Transfer and the like. Folders like luma and gm9 may already exist.

ihaveamac:
Overall I think reducing the amount of files that a user needs to think about helps out both the user (less stuff to copy over and think about what they may have forgotten) and helpers (less figuring out where users placed all the different things)

lilyuwuu:
I'm not seeing what specifically would be easier. The ZIP currently looks like this, which people are told to drag-and-drop to root of SD, merging with existing folders:
image

By packaging the contents of finalize into an archive, we have two main options:

1. Have user create `/gm9/scripts/` and `/luma/payloads`, then have user drag-and-drop `GodMode9.firm` into `/luma/payloads/` and `finalize.gm9` into `/gm9/scripts/` (the location of `finalize.romfs` can be pretty much anywhere). This'll run into the same issues that people do now: people misname `payloads` and/or don't put `GodMode9.firm` into the `payloads` folder (instead putting in the GodMode9 `.zip`, a folder, or an entirely different file).
2. Provide a ZIP that packages `/gm9/scripts/finalize.gm9`, `/luma/payloads/GodMode9.firm`, and `/finalize.gm9` together. This would run into the same issue with merging.

I can see that making it an archive could lessen error (e.g. people somehow dropping the contents of `finalize` to root of SD, or more likely, people interrupting the transfer of the ZIP halfway, resulting in missing files from `SD:/finalize`), but we will still end up with either a problem of [creating folders + dropping files in specific locations] or [incorrectly merging folders].

As-is, the script checks for the existence of `SD:/finalize` at the beginning and halts if it isn't there. It also deletes it only after everything has succeeded (after the NAND backup)

Gruetzig:
you convinced me, but you may want to add github links to the respective tools, so people can check the code, star the repository etc without having to go on a journey

lilyuwuu:
good idea. First thought is to put that in the 'list of homebrew applications installed' thing (the bullet points at the beginning) but I'm a bit iffy on people accidentally clicking them and not knowing where to go on the page. Maybe \<summary> that part (bonus, the page becomes even shorter visually)

Gruetzig:
doubt the accidental link clicking will be a promplem, and if its summary a bit too hidden probably

ihaveamac:
> 1. Have user create `/gm9/scripts/` and `/luma/payloads`, then have user drag-and-drop `GodMode9.firm` into `/luma/payloads/` and `finalize.gm9` into `/gm9/scripts/` (the location of `finalize.romfs` can be pretty much anywhere). This'll run into the same issues that people do now: people misname `payloads` and/or don't put `GodMode9.firm` into the `payloads` folder (instead putting in the GodMode9 `.zip`, a folder, or an entirely different file).
> 2. Provide a ZIP that packages `/gm9/scripts/finalize.gm9`, `/luma/payloads/GodMode9.firm`, and `/finalize.gm9` together. This would run into the same issue with merging.

Another solution to this could be to use a custom GodMode9 build with autoboot but run it like a payload instead of boot.firm. This would cut down the files required into 2 (example `/luma/payloads/AutoSetup.firm` and `/finalize.romfs`). Though this solution wouldn't address the potential typos or misplacement, it cuts it down to one folder hierarchy that could be screwed up.

While I wish we could cut it down to a single file (by including everything in the vram0.tar) the total size is too big for the ~3 MiB limit...

lilyuwuu:
> Another solution to this could be to use a custom GodMode9 build with autoboot but run it like a payload instead of boot.firm. This would cut down the files required into 2 (example `/luma/payloads/AutoSetup.firm` and `/finalize.romfs`). Though this solution wouldn't address the potential typos or misplacement, it cuts it down to one folder hierarchy that could be screwed up.
> 
> While I wish we could cut it down to a single file (by including everything in the vram0.tar) the total size is too big for the ~3 MiB limit...

I'm still a bit iffy on whether it's easier to have people merge everything on the SD card or to have people create the `payloads` folder. Though, if we do it that way, we do avoid people putting in the wrong file (since we'd be linking to the built script rather than GM9. Maybe it's a Windows vs macOS thing? IIRC Windows handles merging just fine but macOS only lets you replace unless you hold option(?)

I'd also rather people have GM9 by the end of the script, so it would need to be retooled so that at the end, the autorunner is deleted, GodMode9 is grabbed from finalize.romfs and placed there, then GodMode9 is copied to ctrnand:/rw/luma/payloads.

The other thing is that this makes maintenance of the script a bit more annoying, since we'd have to build two things to get it to work (the archive, whatever it may be, and the autorun script).

lilyuwuu:
The latest revision of the script fixes a few issues:

- Errors are more clearly indicated, with "Error #XX: (Problem)" on one line. This should make it easier for those who encounter an error when viewing the guide in non-English to figure out what the error is (by visiting Troubleshooting).
- SD card size is now automatically determined rather than relying on the user.
- Dummy title databases are now caught by the script, by hashing the dummy title.db and checking if the hash matches the file (since the hash would have changed once reset). An integrated title.db/import.db would probably still be ideal though, since it would require less user interaction to troubleshoot (just hitting (A) and entering a key combo rather than navigating to a specific location in Settings).

PR #2262, opened 2023-05-18
fix wrong redirection for unsupported kor ssloth-browser

danny8376:
both surprised and not, very specific version of old 3ds kor(/twn/chn) browser, 1.7630/v10240, should be shipped with 11.1~11.8, isn't supported by browserhax. weird gap here, but [it simply exist](https://github.com/zoogie/web/blob/master/nbhax/index.html#L68-L73).
probably will never notice this without [existing case](https://discord.com/channels/196618637950451712/247557068490276864/1108600147878887475)

### DIFF
--- a/assets/js/selecting.js
+++ b/assets/js/selecting.js
@@ -28,15 +28,16 @@ function sslothable(s, n) {
 }
 
 // CVer:NVer map for KOR region
+// KOR/CHN/TWN Old 3DS browser (spider) 1.7630 (v10240, shipped with 11.1~11.8) isn't supported by browserhax
 // CHN/TWN isn't validated for now as those cannot exploit atm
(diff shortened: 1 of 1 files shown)


Gruetzig:
looks like it [also counts for new3ds](https://github.com/zoogie/new-browserhax#whats-needed), unless there were no new models before 11.9

danny8376:
The skater equalievent will be [1.9.10160 (and 1.8.10156, but that's pre-11.4)](https://github.com/zoogie/web/blob/master/nbhax/index.html#L47-L50)
so it should be supported... at least for the first stage.

(also, there's a very specific commit for it [that's only on web repo](https://github.com/zoogie/web/commit/27ec5c7d1128b6b9f69b33f7d68c6c18c7c23902))

PR #2268, opened 2023-05-24
Add system update pages for super-skaterhax and seedminer

lifehackerhansol:
-  EUR/JPN N3DS 11.14-11.15,: super-skaterhax
- TWN 11.15, KOR O3DS 11.15: seedminer


lifehackerhansol:
For now USA is left as "not available"

lifehackerhansol:
We can continue to use seedminer on KOR as usual, no need to change that just yet

lilyuwuu (approved):
lgtm

PR #2269, opened 2023-05-24
add alternate exploits (finish 11.17 guide support)

lilyuwuu:
a.k.a. "the worst structural change in the history of this guide"

War. War never changes. In the year 2019, my great-great grandfather, serving in the army, wondered when he'd get to install CFW on his 3DS for free. He got his wish when zoogie released [Seedminer](https://github.com/zoogie/seedminer) and [BannerBomb3](https://github.com/zoogie/Bannerbomb3), granting free CFW to the vast majority of retail 3DSes in the world.

The Modders awaited Banmageddon; instead, something miraculous happened. We began to use BannerBomb3 not as a weapon, but as a [nearly](https://3ds.hacks.guide/bannerbomb3-fredtool-(twn)) limitless source of [power](https://3ds.hacks.guide/installing-boo9strap-(usm)).

People enjoyed luxuries once thought the realm of science fiction. [Additional free CFW methods](https://github.com/zoogie/unSAFE_MODE/), [the return of an exploit long thought dead](https://github.com/zoogie/new-browserhax), practically [one-click install custom firmware](https://github.com/TuxSH/universal-otherapp). But then, on [March 27th, 2023](https://en-americas-support.nintendo.com/app/answers/detail/a_id/57847/~/wii-u-%26-nintendo-3ds-eshop-discontinuation-q%26a), people awoke from the Nintendo dream.

Years of consumption lead to [shortages of every major exploit](https://twitter.com/mattgcn/status/1660819460859764737). The entire guide unraveled. Simplicity became a [distant memory](https://troubleshoot.uwuu.ca/seventeen). It is still the year 2023. We stand on the brink of [total war](https://wiki.hacks.guide/wiki/3DS:Alternate_Exploits), and I am afraid. For myself, for this guide, for all the people who want to install CFW - because if my time in the army taught me one thing:

It's that war? [War never changes](https://github.com/hacks-guide/Guide_3DS/pull/2269).

---

This PR ties up loose ends caused by 11.17 and brings the guide back to as normal of a state as it can be with BB3 being patched. The following pages are created:

- Seedminer (Mii) for Mii mining, which is the only free way to install CFW on O3DS 11.15
- Alternate Exploits, which are the only way to install CFW on O3DS 11.17
- Seedminer (Alternate), for use with Dumping eShop DSiWare and movable-requiring HBL entrypoints
- Dumping eShop DSiWare, part of the Seedminer + DSiWare + Fredtool sequence
[...]

PR #2270, opened 2023-05-24
add mii mining

lilyuwuu:
♫ Everybody was Mii Seedmining ♫
♪ GPUs fast as lightning ♪
♬ In fact, it was a little bit frightening ♬
♫ But they mined with expert timing ♫

This PR adds mii mining methods to fill a temporary hole left by 11.17. Split from #2269. This involves three new pages:
- Seedminer (Mii)
- BannerBomb3 (11.14 / 11.15)
- Installing boot9strap (Fredtool, 11.14 / 11.15)
  - BB3 and Fredtool's actual webpages are named "bannerbomb3-(legacy)" and "installing-boot9strap-(fredtool-legacy)", respectively

Site navigation has been updated accordingly. JS and TOC changes pending.

Mii mining is proposed to be the automatic method for:
- O3DS on 11.15
- USA N3DS on 11.14, 11.15

### DIFF
--- a/_pages/en_US/bannerbomb3-(legacy).md
+++ b/_pages/en_US/bannerbomb3-(legacy).md
@@ -1,5 +1,5 @@
 ---
-title: "BannerBomb3"
+title: "BannerBomb3 (Legacy)"
 ---
 
 {% include toc title="Table of Contents" %}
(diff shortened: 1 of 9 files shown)


lilyuwuu:
whoops hold on

_pages/en_US/bannerbomb3-(legacy).txt
```diff
@@ -0,0 +1,50 @@
+---
+title: "BannerBomb3 (11.14 / 11.15)"
```
lifehackerhansol, line 2: title is not `BannerBomb3 (Legacy)`

well, it isn't all that important, but we did have an unwritten procedure to have the same title as the file name

lilyuwuu: it's just to make it absolutely clear for the reader because of the risk involved but if we want to keep that convention we can

lifehackerhansol:
rip chromebook users ig

PR #2277, opened 2023-05-27
Fix broken KOR SSLoth-Browser again... actually fix more this time

danny8376:
Well, I'll say never mix integer with bool
This new version literally inverted new/old for kor
I originally wrote it as ``!old`` so new only, it's literally now ``!(new=1=true)`` so old only

Yeah, another [case](https://discord.com/channels/196618637950451712/196635695958196224/1111950239792898128) for me to notice this


lifehackerhansol:
I'm sorry

lifehackerhansol:
merged to HEAD

PR #2278, opened 2023-05-28
Fix missed change for new camera button check

danny8376:
literally just as the title


lilyuwuu:
oh my god why do we keep forgetting this

PR #2293, opened 2023-06-12
troubleshooting: remove redundant SSLoth-Browser cartupdate steps

lifehackerhansol:
This is unnecessary as, if the user used the version selector, they shouldn't be on a cartupdate.


lifehackerhansol:
The goal here was to reduce text in this particular section. However, I'm uncertain about making such an assumption that the user did use the version selector.

Would like opinions before merging or dropping.

lilyuwuu (approved):
ya sure ok

PR #2294, opened 2023-06-12
safecerthax: mii mine if shoulder buttons broken

lifehackerhansol:


_pages/en_US/installing-boot9strap-(safecerthax).txt
```diff
@@ -31,7 +31,7 @@ This exploit will not work on the New 3DS, New 3DS XL, or New 2DS XL. Please ens
 
 {% include_relative include/safemodecheck.txt %}
 
-If the camera did NOT appear, you cannot follow this method. If this is the case, stay on your current firmware and ask [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp) for help.
+If the camera did NOT appear, you cannot follow this method. If this is the case, use [Seedminer (Mii)](seedminer-(mii)) instead.
```
lilyuwuu, line 34: Um... Wouldn't anyone doing safecerthax be anywhere between 11.4 and 11.14? Seedmiiner currently leads to BB3 which won't work as is on 11.4, and anyone with broken shoulders would be doing Fredtool Legacy which is currently only built to handle 11.14 and 11.15.

lilyuwuu: our "why didn't we see this sooner" solution is menuhax67 ->  uloader ... except menuhax only works on 11.7 (U/J) / 11.10 (E) onwards, so not really. woohoooooooo

PR #2297, opened 2023-06-15
troubleshooting: use sensible names in capture title

lifehackerhansol:
- They're all using `compat` which was used elsewhere for technical info boxes
- Not the case here


lifehackerhansol:
nah screw this we can do it later

PR #2299, opened 2023-06-20
Add some more info on kartdlphax

joaoesteves10:
**Description**

This pull request adds a piece of information that might be very helpful for some people trying to get CFW on updated consoles: the source 3DS can be region-changed to match the target 3DS. 

kartdlphax requires two consoles on the same region, but nowhere does the documentation state whether region changed consoles work as the source 3DS; after a desperation move by myself today to get two JPN old 3DS LL on 11.17.0 running CFW, i can confirm it works even if the source 3DS wasn't originally the same region, as I did it using a region-changed EUR old 3DS. 

This info might be very helpful for people in the same situation as me, that have consoles from different regions and don't want to buy a flashcart.

### DIFF
--- a/_pages/en_US/installing-boot9strap-(kartdlphax).txt
+++ b/_pages/en_US/installing-boot9strap-(kartdlphax).txt
@@ -27,7 +27,7 @@ In order to follow these instructions, you will need the following:
 
 - A second 3DS with custom firmware (the **source 3DS**) that is the same region as the 3DS you are trying to modify (the **target 3DS**)
   - The consoles must be USA, JPN, or EUR region consoles
-  - The source 3DS can be one [region changed](region-changing) to match the target 3DS
+  - The source 3DS can be [region changed](region-changing) to match the target 3DS if necessary
 - A physical or digital copy of Mario Kart 7 that is the same region as both consoles
 - An SD card for both consoles
 
--- a/_pages/en_US/installing-boot9strap-(kartdlphax).txt
+++ b/_pages/en_US/installing-boot9strap-(kartdlphax).txt
@@ -27,6 +27,7 @@ In order to follow these instructions, you will need the following:
 
 - A second 3DS with custom firmware (the **source 3DS**) that is the same region as the 3DS you are trying to modify (the **target 3DS**)
   - The consoles must be USA, JPN, or EUR region consoles
+  - The source 3DS can be one [region changed](region-changing) to match the target 3DS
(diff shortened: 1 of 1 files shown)


lifehackerhansol:
On further thought, does kartdlphax actually care about the source console's region?

it **should** have nothing to do with the console's region but rather the *game's* region, i.e. the source console needs the MK7 game from the target console's region.

Correct me if I'm wrong, because at the moment my 3DS is across the Pacific ocean

joaoesteves10:
I had the same thought, but sadly it seems like it does: before region changing the european one I tried using it with the JPN MK7 and the game didn't show up in the target (japanese) consoles, so i'm guessing download play has region locking built in and won't show games from consoles from other regions, even though the game's region matches the target console's

danny8376:
> On further thought, does kartdlphax actually care about the source console's region?
> 
> It **should** have nothing to do with the console's region but rather the _game's_ region, i.e. the source console needs the MK7 game from the target console's region.
> 
> Correct me if I'm wrong, because at the moment my 3DS is across the Pacific ocean and I can't test anything.

Afaik, console region do affect, the reason there is that 3ds download play itself is region locked (based on console region, regardless of game design), hence luma have patch to unlock it (only work when both modded afaik)

_pages/en_US/installing-boot9strap-(kartdlphax).txt
```diff
@@ -27,6 +27,7 @@ In order to follow these instructions, you will need the following:
 
 - A second 3DS with custom firmware (the **source 3DS**) that is the same region as the 3DS you are trying to modify (the **target 3DS**)
   - The consoles must be USA, JPN, or EUR region consoles
+  - The source 3DS can be one [region changed](region-changing) to match the target 3DS
```
lifehackerhansol, line 30: `can be one` -> `can be`

If you look the other way, both are grammatically correct, but I feel that it should be worded so that "you can do this to make kartdlphax work" instead of "this one can be one that did that work so that kartdlphax works", if that makes sense?

lifehackerhansol:
Seems fine outside of the above.

I guess I'm a little out of touch; I feel that just mentioning two consoles with identical regions would imply that a region change works but I suppose not everyone understands that exactly.

_pages/en_US/installing-boot9strap-(kartdlphax).txt
lilyuwuu: +1

PR #2301, opened 2023-06-24
post-safemodecheck: clarify that just the camera app needs to work

lifehackerhansol:
- It is not uncommon for users to have a broken camera
- In this case, camera app will try to load and then crash, which implies that shoulder buttons do work
- Let's make life easier for these users


lilyuwuu:
Couldn't we just word this as "If the screen changes" or something? I feel like people will see the but crashed thing and then think that implies that it won't work

lifehackerhansol:
stale

PR #2308, opened 2023-07-07
re: ~~ctrtransfer/~~region-changing: add faketik

danny8376:
do lily forget to git add _pages/en_US/region-changing.txt?


Gruetzig:
whoever doesnt do git add . is sus

PR #2313, opened 2023-07-13
troubleshooting: initial frogtool steps

lifehackerhansol:
**Description**

Most of this is a straight-copy from fredtool. Only the DSiWare steps are added; ideally there should be more troubleshooting steps for other sections as well.

Feedback would be appreciated either for issues with this PR or for adding more troubleshooting.


lifehackerhansol:
stale

PR #2314, opened 2023-07-13
troubleshooting: move universal-otherapp SB9SI error to "Issues with SafeB9SInstaller"

lifehackerhansol:
**Description**

As per suggestion that a less-tech-inclined user could potentially be looking in the wrong place.


lifehackerhansol:
Need some ideas for header... it looks weird without one, but the one I wrote isn't that great either

lilyuwuu (approved):
ya sure

PR #2315, opened 2023-07-14
Version validation

danny8376:
* add version validator to prevent ppl from entering version like 1.17
* blacklist 11.17 for K/C/T


Gruetzig:
wow so cool :)

PR #2318, opened 2023-07-18
Add CTRNAND.

TheMasterCoders:
Added CTRNAND. Somebody translate it and fix it

**Description**

It just adds how to add the updated Luma3DS version to CTRNAND. Only in en-US


ihaveamac:
This page does not assume the console has GodMode9 set up. Additionally, recent versions of Luma3DS will copy itself to CTRNAND (though only once per update).

lifehackerhansol:
As mentioned above, since Luma3DS 11.0, Luma3DS copies itself to CTRNAND. So this is unnecessary.

PR #2320, opened 2023-07-19
lets screw this legacy shit

Gruetzig:
theres absolutely no reason other than "AHUJAFSGOI/A(&TUASDT%ASOD(TSASOTUS"


lifehackerhansol (changes requested):
If we're going to nuke these pages, then inline the relevant includes to its respective pages, since they're no longer in need of consolidation.

lifehackerhansol (approved):

PR #2322, opened 2023-07-19
add a redirect for pages we've removed

lilyuwuu:
does what it says on the tin


lifehackerhansol:
We could consider redirecting `homebrew-launcher-*` to its `installing-boot9strap-*` counterparts; in both cases, it's the primary exploit, so it doesn't really break anything.

PR #2325, opened 2023-07-19
Prioritize super-skaterhax over Seedminer for N3DS 11.16

### LINKED ISSUE #2323: [Enhancement] Consider recommending super-skaterhax over bb3 for N3DS 11.16
With nimdsphax now being a thing, I do believe it would be fair to consider that the currently recommended pathway for N3DS 11.16 could lead to overall more confusion (or run into more issues) than super-skaterhax.

Right now, the pathway for N3DS 11.16 currently is:
Seedminer -> bb3 -> menuhax67 -> nimdsphax

This has several issues that could come up:
- The system could be unmineable (which would make this not a viable pathway, and while unmineable’s do eventually get recommended super-skaterhax anyways, why do this extra effort)
- The system could have potentially had CFW uninstalled without removing present DSiWare, which would break the DSiWare Management menu (and by extension, bb3)
- While not necessarily an actual issue (and more of a user error case), the extra steps of removing menuhax67 could be missed (this absolutely can happen), and regardless, it’s overall more complex to install and then uninstall menuhax67 rather than just entering into HBL once via the browser

Yes, super-skaterhax does have pretty bad reliability, but imo the bad reliability is a more than acceptable trade-off (especially now given nimdsphax) given the many more potential “failure points” or “user error points” that one could run into compared to super-skaterhax

TheMasterOfMike:
**Description**
~~hopeful I didn't break the js but I wouldn't be shocked if I somehow managed to do so~~

As per the various reasons described in #2323 (that I won't restate here for brevity sake), there are valid reasons (especially now with nimdsphax) to recommend and prioritize super-skaterhax over Seedminer on N3DS 11.16.

This PR:
- Gives priority to super-skaterhax over Seedminer for N3DS 11.16
- Links to "Updating Firmware (New 3ds)" instead of "Updating Firmware (KOR)" for KOR N3DS's on 11.14/11.15 (due to the latter linking to Seedminer in comparison to the former linking to super-skaterhax)

Closes #2323

### DIFF
--- a/assets/js/selecting.js
+++ b/assets/js/selecting.js
@@ -125,8 +125,8 @@ function can_safecerthax(major, minor, native, region, model) {
 }
 
 // super-skaterhax
-// EUR/JPN/USA: 11.17 only
-// KOR: 11.16 only, KOR does not have 11.17 (Do seedminer first though)
+// EUR/JPN/USA: 11.16 and 11.17
+// KOR: 11.16 only, KOR does not have 11.17
 // CHN/TWN has no N3DS
 function can_superskaterhax(major, minor, native, region, model) {
     let do_redirect_sysupdate = false;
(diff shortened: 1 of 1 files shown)


lifehackerhansol:
Superseded by b8898b77cd5daa9c5877f5673f0aac5f6bed8b1f.

danny8376:
If you really prefer super-skaterhax, don't let them update and try super-skaterhax directly, it doesn't ask update on 11.16 iirc
We did prioritise super-skaterhax over seedminer on 11.16 at some point (before 11.17 released), then we make it fallback only later
It's simply not as stable as bb3, ppl can spend hours on tweaking various things to try to make it work (and we don't even really sure what can make it work sometimes), while bb3 can just be done stably

PR #2328, opened 2023-07-20
pichaxx sd image update

corduroy1:
**Description**

this as well yay woo


lilyuwuu:
hansol can probably fix this for you if he wants but generally it'll be easier if you make master even with upstream and make changes on a separate patch branch so that you don't have to make extra merge commits based on changes that have already happened

corduroy1:
realized this right after sending lol still learning oops. thanks :)

lifehackerhansol:
This pull request is wrong anyway. It's actually not the author's fault, but rather mine, because we don't need movable.sed on root anymore.

Interestingly, this screenshot was actually never updated for _**11.16**_, so we can just reuse the existing one lol

lilyuwuu:
does it help reliability with nimdsphax tho?

lifehackerhansol:
There is no mention of a movable.sed anywhere in nimdsphax's source code.

lifehackerhansol:
Actually, the original screenshot still needs an update, as it lacks a 3ds folder needed for nimdsphax.

I'm going to write up a proper issue tracker for this.

PR #2336, opened 2023-07-22
treewide: standardize bullet points

lifehackerhansol:
**Description**

This standardizes bullet points across the board:
+ What You Need: *
+ All others: +

Without this, managing strings has become increasingly difficult, especially on the side of translations as even these differences cause Crowdin to detect it as separate strings.

It also actually keeps a synchronized format so it looks cleaner when editing.


lilyuwuu (approved):
ya sure

PR #2340, opened 2023-07-23
TEMP: use 12.0.1 in a few more pages, update to latest luma in finalize

lilyuwuu:
Change recommended for standardization purposes (currently soundhax and ntrboot use 12.0.1 and don't prompt people to update which is inconsistent) At the moment, 13.x is unable to boot on at least 10.6 and below.  So, to ensure that guide pages are (at least mostly) consistent, using 12.0.1 on all* pages and then prompting user to update luma in finalizing setup. This involves reorganizing finalizing setup a bit so that the cfw check and system update come before we modify the sd card.

Note: Change currently not copied to installing-boot9strap-(kartdlphax) as this would involve reverting to an older commit and overall complicate the instructions. Also not copied to restoring-updating-cfw as people may be in scenarios where they don't continue onto finalizing setup.


lilyuwuu:
The other option to this is to keep luma 12.0.1 for just soundhax and ntrboot (and i guess hardmod too in theory, as if people actually do that). But this continues our current inconsistency between guide version usage (not that this isn't also kind of a problem w/ kartdlphax) and would mean that a lot of people (namely, anyone following anything other than soundhax/ntrboot/hardmod) would be redundantly updating luma and probably questioning why we make people do that.

lifehackerhansol (changes requested):
- wording on a2b page suggests users to copy boot.firm from the latest version which isn't the case

seems ok otherwise?

lifehackerhansol (approved):

PR #2343, opened 2023-08-09
Changed latest to 12.0.1 in troubleshooting

Botman836:
I pointed it out, lily said cool so i decided to push it myself :thumbsupemoji:


lilyuwuu:
I had figured this is rolled together with firm patches error which has lower minimum firmware & could theoretically occur on even older fw (e.g. Arm11 exception handlers are enabled on Luma <=7.1). Though upon further reflection there don't actually seem to be any common scenarios where this could occur? The other logical issue here is that people who read this line would never get prompted to update Luma back to latest after updating their console, which we want to avoid

lifehackerhansol:
With Luma3DS 13.0.2, this pull request is moot.

PR #2345, opened 2023-08-10
get-started: switch from CSS-based error messages to alerts

lifehackerhansol:
May make it more explicit that user entered something wrong. Confirmed working with translations. No idea if it works on all browsers.


lifehackerhansol:
This doesn't seem like a popular idea

PR #2346, opened 2023-08-12
Page for petit-compwner exploit

TophattedWasTaken:
This may be unnecessary, considering it is just one DSiWare game that likely not many people own, and they can use other (albeit slower) exploits to achieve the same results. Regardless, I made a page for it anyways. If you think it isn't useful enough to add, that is alright too. If you find any errors I made, feel free to fix them. I want this guide to be as helpful as possible.


lilyuwuu:
I don't personally see anything wrong with the PR itself (aside from very small nit-picky stylistic stuff), but as you point out it is indeed quite niche. Though it does admittedly save quite a bit of time over Seedminer+Fredtool, for the (also admittedly few) people that have this game.

@lifehackerhansol thoughts?

TophattedWasTaken:
Just note that I haven't been able to test this yet due to me not having another 3ds. I don't see why it would be patched though, considering it is an obscure exploit that from my limited understanding has to do with DS mode and not NATIVE_FIRM. Worst case scenario, b9sTool doesn't boot.

lifehackerhansol:
The idea isn't awful. But I don't know about whether it belongs in the guide, per se. (what with our recent cutting down on using b9stool as much as possible)

Think this just needs some marinating for a bit.

danny8376:
Since this is DSiWare so they'll use fredtool anyway and we also have smilehax, which is equally niche imo, I don't see why we can't also put this in the guide, but not sure if it's really worth it as they can just use fredtool and we can have less confusing-ish (but already confusing enough to some people) alternate page.

TophattedWasTaken:
Whatever you guys think is best. I don't want to make the guide more confusing.

TophattedWasTaken:
About to test the exploit to see if it's patched or not. I can't launch b9sTool on my cfw 3ds since I don't want it to get bricked, so I'll use the DSi homebrew launcher. Should achieve the same end result of figuring out if the exploit has been patched.

Gruetzig:
you wont get patched as long as you dont select install boot9strap

TophattedWasTaken:
That's good to hear, already installed the DSi hbmenu files however.

TophattedWasTaken:
It works! (in other news...how do you close the DSi homebrew menu?)

lifehackerhansol:
Site needs to be added to site-navigation page and the progress bar in `_main.js`; if you aren't comfortable with javascript, I can handle the last part, but please do the first part.

TophattedWasTaken:
I don't know JavaScript, adding it to site navigation right now

lifehackerhansol (approved):

PR #2351, opened 2023-08-16
Update installing-boot9strap-(soundhax).txt

TophattedWasTaken:
Exploit pages use the exploit name for the section that launches SafeB9SInstaller (from what I've seen, at least)


lilyuwuu:
IIRC "Launching SafeB9SInstaller" is the original text for these sorts of methods. But consistency was probably lost at some point.

Will have to look into what consistency would be in this case for the other exploits and then change them accordingly (but using "Launching SafeB9SInstaller" for all of them would probably be better because it means fewer guide translation strings)

lifehackerhansol:
+1 on moving all other text to `Launching SafeB9SInstaller`

TophattedWasTaken:
Consistency is good.

PR #2357, opened 2023-08-19
Add frankenfirm soundhax to troubleshooting

### LINKED ISSUE #2279: Providing cartupdated version crossing two releases can lead to useless Soundhax guide
**Pages with issue(s)**

https://3ds.hacks.guide/get-started

**Description of the issue(s)**

Providing a version like `7.1.0-0E` will lead to soundhax guide that may not work (it's complicated, as some cart update partitions provided Nintendo 3DS Sounds updates, and some didn't).

I think the guide should be either updated to handle those versions (there is a special build specifically for those versions - https://github.com/nedwill/soundhax/pull/114), or soundhax shouldn't be suggested when NVer is between 0 and 4 and CVer is at least 5.0.0 (indicating cart update).

danny8376:
first, ref: https://gist.github.com/danny8376/524394a41202c3617e6a1ef6c181fa91

the site should cover all cases include 2.1.0 handling, hopefully
[js part of the site](https://github.com/nedwill/soundhax/compare/master...danny8376:soundhax:frankenfirmware#diff-4d6443d6dc5d62f22ddb20623fe3686e9df0b374acbfed8fe67e8feba447f911)

and this should resolve #2279
i don't think there's a good way to determine which sound app version they have without running normal soundhax first, even though it's almost guaranteed to fail (for 5.0.0 and greater)

### DIFF
--- a/docs/public/assets/js/link-common.js
+++ b/docs/public/assets/js/link-common.js
@@ -28,6 +28,11 @@ if (selectedVersion && window.MATCH && window.generateLink) {
     } else {
         replaceLink();
     }
+
+    if (!window.autoLinkAddedCustomEvent) {
+        window.autoLinkAddedCustomEvent = true;
+        document.addEventListener("customEventReplaceLink", replaceLink);
+    }
 }
 
 })();
--- a/docs/public/assets/js/soundhax-frankenfirm-link.js
+++ b/docs/public/assets/js/soundhax-frankenfirm-link.js
@@ -1,0 +1,145 @@
+/*
+    Copyright (C) 2024 DannyAAM
+
+    SPDX-License-Identifier: MIT
+*/
+
+(() => {
+
+const BASE = "https://github.com/danny8376/soundhax/raw/frankenfirmware/soundhax";
+
+window.MATCH = "soundhax.686178.xyz/frankenfirm";
+
+// adopt from https://github.com/danny8376/soundhax/blob/2a2c2499ad45d9216ae4c9d157dfb90dfd3b6bd4/docs/js/frankenfirm.js#L82-L237
+window.generateLink = (major, minor, nver, region, model, link) => {
+    console.log("MATCH")
+    const fileRegion = FILENAME_REGION_MAP[region];
+    const errorIndex = link.lastIndexOf("?");
+    if (errorIndex == -1) {
+        return null;
+    }
+    const error = link.slice(errorIndex + 1);
+    console.log(error);
+
+    if (model == DEVICE_N3DS || ["K", "C", "T"].includes(region)) {
+        return null;
+    }
+
+    switch (parseInt(nver)) {
+        case 0:
+        case 1:
+        case 2:
+        case 3:
+            console.log("nver 0-3")
+            if (major < 2 || (major == 2 && minor <= 1)) {
+                switch (error) {
+                    case "unplayable":
+                        // very unlikely to have v1027 sound app in this case?
+                        return `${BASE}-${fileRegion}-o3ds-pre2.1.m4a`;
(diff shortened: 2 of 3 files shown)


_pages/en_US/troubleshooting.txt
```diff
@@ -276,13 +276,19 @@ If your console is not on those firmwares, it likely indicates that you already
 <summary><u>"An error has occurred, forcing the software to close..." (white message box)</u></summary>
 
 There is an issue with your `otherapp.bin` file (it is missing, misplaced, or corrupted). Download the latest release of [universal-otherapp](https://github.com/TuxSH/universal-otherapp/releases/latest) and place it on the root of your SD card.
+
+If the above doesn't fix your issue, and you are using an Old 3DS / Old 3DS XL / Old 2DS, you may be encountering a specific issue involving cartridge updates. Take another look at the system version. If the last number, e.g. 4 in 11.3.0-**4**U, is 4 and lower and the system version is 11.3.0 and lower, replace the soundhax file on the root of your SD card with the one from [this site](http://soundhax.saru.moe/frankenfirm.html?crash)
```
lilyuwuu, line 280: "specific" is not necessary here
4 and lower -> 4 or lower
soundhax should be capitalized
"If the last number is 4 or lower (e.g. 11.3.0-**4**U) would be preferred

Add link to NH here too? And or maybe to safecerthax?

_pages/en_US/troubleshooting.txt
```diff
@@ -276,13 +276,19 @@ If your console is not on those firmwares, it likely indicates that you already
...
 {% endcapture %}
 <details>{{ compat | markdownify }}</details>
 
 {% capture compat %}
 <summary><u>"Could not play"</u></summary>
 
-You have the wrong Soundhax file for your console and region, or your console is incompatible with Soundhax. In the latter case, your course of action will determine on what version your 3DS is currently on. Join [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp) for assistance.
+You have the wrong Soundhax file for your console, region and version. Double-check that you entered all the info correctly into the [Soundhax website](https://soundhax.com/), redownload your soundhax file, and replace it with the one on the root of your SD card.
```
lilyuwuu, line 287: Soundhax should be capitalized

_pages/en_US/troubleshooting.txt
```diff
@@ -276,13 +276,19 @@ If your console is not on those firmwares, it likely indicates that you already
...
+
+If the above doesn't fix your issue, and you are using an Old 3DS / Old 3DS XL / Old 2DS, you may be encountering a specific issue involving cartridge updates. Take another look at the system version. If the last number, e.g. 0 in 11.3.0-**0**U, is 3 and lower and the system version is 11.3.0 and lower, replace the soundhax file on the root of your SD card with the one from [this site](http://soundhax.saru.moe/frankenfirm.html?unplayable)
```
lilyuwuu, line 289: see other review referencing this text
Link to safecerthax here if troubleshooting fail?

_pages/en_US/troubleshooting.txt
danny8376: should ``11.3.0 and lower`` also be ``11.3.0 or lower``? (Actually, this part might not even be needed as they shouldn't go soundhax if not?)
not sure about safecerthax, indeed could be an alternative, simply add another section of ``If the above doesn't fix your issue, and you are using an Old 3DS / Old 3DS XL / Old 2DS, you can also try...``?

Gruetzig (approved):

danny8376:
This PR is currently based on #2498

lifehackerhansol (approved):

PR #2361, opened 2023-08-31
Site Navigation Adding Missing Pages

TophattedWasTaken:


Gruetzig:
do we want those pages in site nav though? people shouldnt come across them without selector imo, and theres no benefit to going to updating pages via site nav?

PR #2373, opened 2023-10-05
Add MSET9

lilyuwuu:
Huzzah, the drought ends! Long live 3ds hax! (Hopefully nothing drastic will happen [anytime soon](https://twitter.com/NintendoAmerica/status/1709418623566020975)...)

---

This PR adds the MSET9 exploit, adapted from the [wiki page](https://wiki.hacks.guide/wiki/3DS:Alternate_Exploits/Installing_boot9strap_(MSET9)) graciously written by @Gruetzig.

- Replaces Alternate Exploits in guide routing with MSET9
    - Link to Alternate Exploits if MSET9 cannot be performed (Mac, mobile users)
- Update FAQ and 11.17 notice to account for us now having frii hax on 11.17 o3ds

TBD:
- ~~Did I break stuff in attempting to make JS changes?~~ Apparently not, somehow!
- Does MSET9 really work on CHN and TWN? If so we may want to route CHN TWN 11.16 to this method too. (I dunno, is MSET9 easier than Bannerbomb3-Fredtool (TWN)?)
- ~~Add root of SD (and command prompt?) screenshots~~ done!
- ~~oops i forgot to add this to site-navigation lmao~~ fixed


Gruetzig:
merge if you have balls

TheMasterOfMike:
Few things:
- Technically, Mac users could run Linux freely (be it through live booting or through a VM - though I'll conceed that it's probably insanely impractical for Apple Silicon Macs)
- alternate-exploits should probably have its start changed since it currently says "3DS / 3DS XL / 2DS consoles on firmware 11.17.0 do not currently have any free, software-only methods." which, given this, is not exactly true anymore
- It may be worth waiting even longer on this, since (among other things, such as proposed file location changes), based on the MSET9 readme, it says "(new3ds will be coming at some point)"

I do think there is also reason to probably consider recommending it over stuff like old-browserhax+SSLoth, safecerthax, Mii Mining, and any forms of bb3 (for 11.8-11.17) (disclaimer: I have not had a chance to look into stuff like success rate on MSET9, but it looks like a case of "if steps are followed correctly it should be almost 100%") (mentioning this as well since there are also js changes being made here so why not do all of them at once):
- safecerthax requires working L+R+D-Pad Up+A buttons, plus an internet connection
- old-browserhax+SSLoth requires an internet connection (though in fairness, this is probably the least worth recommending MSET9 over)
- Mii Mining is... Mii Mining (literally anything that doesn't require external hardware/software will be better than Mii Mining)
- bb3 has stuff like a moveable being unmineable, DSiWare Management being broken because somebody uninstalled CFW, and not to mention (in the case of TWN should it actually work there) anything like this should be better than Fredtool anyways

lilyuwuu:
I generally don't have any issues with what you've said this time ;) you're bringing really good points I didn't think about.

- There's actually been some theory crafting for macOS to make it less impractical, we shall see.
- Good point about Alternate Exploits, generally forgot XP There might be some other 11.17 o3ds remnants elsewhere too.
- General consensus internally (and by consensus I mean @lifehackerhansol) is to wait a bit longer anyway, that's part of why this is a draft. Stability thus far has generally been fine (I haven't noticed any data loss or unfixable issues yet), but the actual instructions can indeed still change during the beta phase. (I'm generally erring towards merging sooner rather than later because I feel O3DS being moddable for free with this method is more important; it's not ideal for people to unnecessarily be buying things)
- You wouldn't have any reason to know this, but Mii mining actually has a BFM-style service that's seemingly been working well from testing on NH so far. (And Mii mining will be mandatory for all Seedminer routes anyway once online play goes down, since that almost certainly includes friend servers.)
- I'm otherwise hesitant to recommend this over safecerthax, SSLoth, or BB3 most critically because this method REQUIRES a proper computer and a good chunk of people use other devices (organization managed / locked down Chromebooks, phones, tablets). I do anticipate additional JS changes if we confirm this works on CHN though (kind of on the fence for TWN due to the special method we have to maintain for it)

TheMasterOfMike:
> * General consensus internally (and by consensus I mean @lifehackerhansol) is to wait a bit longer anyway, that's part of why this is a draft. Stability thus far has generally been fine (I haven't noticed any data loss or unfixable issues yet), but the actual instructions can indeed still change during the beta phase. (I'm generally erring towards merging sooner rather than later because I feel O3DS being moddable for free with this method is more important; it's not ideal for people to unnecessarily be buying things)
That's reasonable - the main thing I was thinking of is that "why do two major changes to the js when we can do two at the same time" but that logic also makes sense
> * You wouldn't have any reason to know this, but Mii mining actually has a BFM-style service that's seemingly been working well from testing on NH so far. (And Mii mining will be mandatory for all Seedminer routes anyway once online play goes down, since that almost certainly includes friend servers.)
Wasn't considering that killing online play servers would also kill friend servers but this actually makes sense (I still don't understand why Nintendo is doing this but that's an entirely separate and unrelated discussion altogether) - that should alleviate a lot of the pain that comes from Mii Mining at the moment, which is nice
> * I'm otherwise hesitant to recommend this over safecerthax, SSLoth, or BB3 most critically because this method REQUIRES a proper computer and a good chunk of people use other devices (organization managed / locked down Chromebooks, phones, tablets).
Honestly imo I feel like we're overestimating the number of people who fall in that group, sure, it's not zero, but keep in mind:
- A lot of modern phones/tablets don't even offer an SD Card slot in the first place
- Of those that do offer them, a lot of them are microSD card slots, which means if the person is still using the stock SD Card (which is a full size SD Card on O3DS) they're still going to have issues anyways _regardless_ of whether or not they were using MSET9

While I do agree that it actually is probably not worth recommending this over old-browserhax+SSLoth (due to the only real hurdle that exists for it being an internet connection, which is much less likely to be a problem than the lack of a proper computer), I still think that it's far more likely that safecerthax and any mining (Mii Mining or Seedminer) -> bb3 pathway have a higher chance of hitting a hurdle (e.g. broken buttons, unmineable moveable, DSiWare Management being broken, etc.) than the chance that MSET9 hits a hurdle because they don't have a proper computer

lilyuwuu:
Good point! I know you wouldn't have access to this either, but our analytics tell us that about a little over half of our views come from mobile (and the other half of course from desktop). Generally midrange to low end phones still have microSD card slots too, and I'd also say it's probably a bit easier to find a variety of microSD cards in a random store than full size SD cards these days.

TL;DR: The mobile age is the future and [I must scream](https://en.wikipedia.org/wiki/I_Have_No_Mouth,_and_I_Must_Scream)

TheMasterOfMike:
> Good point! I know you wouldn't have access to this either, but our analytics tell us that about a little over half of our views come from mobile (and the other half of course from desktop). Generally midrange to low end phones still have microSD card slots too, and I'd also say it's probably a bit easier to find a variety of microSD cards in a random store than full size SD cards these days.
> 
> TL;DR: The mobile age is the future and [I must scream](https://en.wikipedia.org/wiki/I_Have_No_Mouth,_and_I_Must_Scream)

Honestly in an ideal world, it would be possible to just have MSET9 as the primary for 11.8-11.17 O3DS, with the others being linked as alternate options, but I don't think there's a good way to do that (outside of maybe reworking the selecting.js file to mention the pre-requisites but that would both be a lot of effort and is also _**way**_ out of scope for this

lifehackerhansol:
Replacing 11.8-11.13 with MSET9 is a no from me for the same reasons lily has mentioned above. (Though there's something to be said for 11.14-11.15 as, even with the whole service thing being beta-tested within the server, Mii mining is just another 2 hours of sitting around hoping for magic.)

All of this will be decided at a future date when this exploit gets off beta, until then how about we just rejoice at the fact that we have a free exploit on O3DS again? We're jumping the gun too early here. (Why is everyone in such a rush lately?)

_pages/en_US/installing-boot9strap-(mset9).txt
```diff
@@ -0,0 +1,104 @@
...
+	+ You should keep this window open until the end of this page, but if you close it by accident, you can open it again by redoing this step
+
+	![Image: MSET9 setup](/images/screenshots/mset9/mset9-setup.png)
+	{: .notice--info}
+
+1. Type `1`, then press Enter to setup MSET9
+	+ If you see "(title.db/import.db) does not exist on SD card!"  and "Create them now? (type yes/no)", follow the [troubleshooting guide](troubleshooting#installing-boot9strap-mset9)
+1. Once the window says "Setting up... done", safely eject your SD card from your computer
+	+ Keep the MSET9 window open
+1. Insert the SD card into your console
```
lifehackerhansol, line 59: Sync this text with other guide pages (maybe?)

_pages/en_US/installing-boot9strap-(mset9).txt
```diff
@@ -0,0 +1,104 @@
...
+{: .notice--danger}
+
+1. **[Hover over](/images/screenshots/mset9/hover-settings.png)** the System Settings icon using the D-Pad (do not select it yet)
+1. Power off, then power on your console
+1. Press (A) to launch System Settings
+1. Navigate to `Data Management` -> `Nintendo 3DS` -> `Extra Data` ([image](/images/screenshots/mset9/settings-extdata.png))
+1. **Do not press any buttons or touch the screen** - ensure that you see the Mii Maker icon
+1. **With the console STILL ON, and without pressing any buttons or touching the screen**, remove the SD card from your console
+	+ The menu will refresh and say that no SD card is inserted, which is expected
+1. Insert the SD card into your computer
```
lifehackerhansol, line 74: s/the/your/
sync with other pages

_pages/en_US/installing-boot9strap-(mset9).txt
```diff
@@ -0,0 +1,104 @@
...
+
+This method currently only works on Old3DS-family consoles (3DS, 3DS XL, 2DS) on firmwares between (and including) 11.8.0 and 11.17.0.
+{: .notice--warning}
+
+This method requires a computer running Windows or Linux. macOS is NOT supported, nor are phones or tablets. If you do not have access to Windows or Linux, you will need to use an [alternate exploit](alternate-exploits).
+{: .notice--warning}
+
+### What You Need
+
+* The latest prerelease of [MSET9](https://github.com/zoogie/MSET9/releases) (the Release `.zip`)
```
lifehackerhansol, line 25:
```suggestion
* The latest prerelease of [MSET9](https://github.com/zoogie/MSET9/releases) (the Release `.zip` file)
```

lifehackerhansol (changes requested):
For all .txt files, replace tabs with spaces

lifehackerhansol:
> For all .txt files, replace tabs with spaces

4-space indent to be specific

_pages/en_US/installing-boot9strap-(mset9).txt
```diff
@@ -0,0 +1,104 @@
...
+1. Copy everything inside the `SDMC` folder from the Release `.zip` to the root of the SD card, overwriting any existing files
+
+	![MSET9 root layout](/images/screenshots/mset9/mset9-root-layout.png)
+	{: .notice--info}
+
+1. Execute the `mset9.py` script on the root of the SD card
+	+ **Windows**: Double-click `mset9.bat`
+		+ If this doesn't work: hold SHIFT and right-click anywhere in the SD card window, then click the option that opens Terminal, PowerShell, or Command Prompt, then type `py -3 mset9.py` and press Enter
+	+ **Linux**: open a Terminal window, `cd` to the root of your SD card, then type `python3 mset9.py` and press Enter
+	+ *You should see a window that says "MSET9 ... SETUP by zoogie"
```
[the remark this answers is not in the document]
lilyuwuu: it's what happens when you try to find and replace things and then forget about wiki markup

lifehackerhansol (approved):

lifehackerhansol (approved):

PR #2385, opened 2023-10-18
mset9: changes for 1.1

lilyuwuu:
- MSET9 now has extensive error-checking, which is almost fully accounted for in the troubleshooting (save for errors that shouldn't happen)
- Enhanced troubleshooting for ID1, pulled from the visual troubleshooter. Commonized and used for Seedminer multiple ID1 scenarios
- Update MSET9 page to account for mset9 v1.1
- Finalizing script v1.5.1 to remove errors.txt


_pages/en_US/troubleshooting.txt
```diff
@@ -463,15 +520,19 @@ Once you've done this, continue from [Section I Step 7](installing-boot9strap-(m
 <details>{{ compat | markdownify }}</details>
 
 {% capture compat %}
-<summary><u>(title/import).db does not exist, create them now?" / invalid database</u></summary>
+<summary><u>Error 08: One or more files are missing or malformed!</u></summary>
+
+One or more files that MSET9 needs to run is missing or corrupted. Re-download the [MSET9 ZIP](https://github.com/zoogie/MSET9/releases/latest) and extract it to the root of your SD card, replacing all existing files, then try again.
```
lifehackerhansol, line 525: not super important but ZIP is capitalized

_pages/en_US/troubleshooting.txt
```diff
@@ -556,27 +582,20 @@ You most likely did something different from the MSET9 instructions, or selected
     + You should see a window that says "MSET9 ... SETUP by zoogie"
 1. Type the number corresponding to your console model and version, then press Enter
 1. Type `3`, then press Enter
-1. Once the window says "Deleting... done.", insert the SD card into your console
+1. Once the window says "Successfully removed MSET9!", insert the SD card into your console
```
lifehackerhansol, line 585: reinsert SD?

_pages/en_US/troubleshooting.txt
lilyuwuu: imo file extensions should always be capitalized

lifehackerhansol: we always used `.zip` for that

on that note, we should also explicitly specify _not_ to get the source code which is also a ZIP

lifehackerhansol (approved):

PR #2393, opened 2023-11-09
Ultimate method cleanup because MSET9

Gruetzig:
* Replace Seedminer page with dummy page that tells user about MSET9 and directs to index page
* Cleanup includes and troubleshooting
* Remove all alternate exploits, seedminer, bannerbomb3, menuhax, twn exclusive pages


lilyuwuu:
shouldn't we just have it only be table if our routing is this simple now

Gruetzig:
Its quite fancy still and we can prevent people from misreading 11 as 1

lifehackerhansol:
Can we keep browserhax and safecerthax

I understand nuking Mii mining et al but I can't think of a good argument as to why we remove those two

Gruetzig:
internet access, its browserhax, safecerthax uses safe mode, whats your problem with mset9 exactly?

lifehackerhansol:
I'm thinking difficulty level here. Yes, MSET9 is easy enough now, but also browserhax/safecerthax is ridiculously easy to do.
Hell, when I brought these back, it was because _Seedminer_ was annoying, and this was back when BB3 was actually functional.

I can see not using safecerthax with safe mode and all, but browserhax is still pretty much a one click hack. If that's not a good enough reason, then sure, I guess.

Gruetzig:
if there was no proxy setup and no odd shenanigans on japanese new model 3ds consoles, sure id agree with one click hack, but currently it really isnt

lilyuwuu:
I'm also of the opinion that keeping browserhax and safecerthax would probably be better for now since those are easier to troubleshoot. What I specifically dislike about menuhax/kartminer/super-skaterhax is that a lot of our troubleshooting effectively amounts to throwing things at the problem until it gets fixed, which doesn't happen as much with ssloth browser or safecerthax

Also if someone doesn't have internet access i feel like they're gonna have bigger problems than not being able to do these methods

Gruetzig:
You really want people to set a custom proxy huh? Yes, SSLoth-Browser and safecerthax seem relatively stable (maybe because you dont see the method being used a lot in assistance), but MSET9 isn't worse, right?  I really dont like that a proxy has to be set. And theres nothing that speaks against it, its not like removing it from the guide eliminates the methods forever, for the niche niche cases we can direct people to the wiki, MSET9 is just better... :((. 

Of course a dead WIFI chip is a rarity, and incompatible routers  (cough cough 5g) too, that wasn't my main point, but MSET9 is offline, which even if only slightly still makes a difference. 

What makes it better than MSET9 that deserves it to stay on the guide?

lifehackerhansol:
There is now no proxy for browserhax (which should resolve that weird error code we kept getting): https://github.com/hacks-guide/Guide_3DS/pull/2394

Thank me later

lilyuwuu:
image

TheMasterOfMike:
honestly I also agree with lifehackerhansol and lilyuwuu as well, with the reasonable solution methodology to laying everything out being:
- 1.0-11.3: soundhax
- 11.4-11.13: ssloth+nbh/obh (if bad ISP or dead WiFi, direct to mset9)
- O3DS 11.14: safecerthax (if bad buttons, bad ISP, or dead WiFi, direct to mset9)
- N3DS 11.14: mset9
- 11.15-11.17: mset9

Back when MSET9 was originally being discussed, I myself had brought up universally doing MSET9 for 11.4+, and while back then it was initially dismissed only because of the platform compatibility issue (and also because it was new I guess but like whatever) but thinking about it more since then, SSLoth and safecerthax are overall easier exploits for an average person to use than MSET9, so _why not keep them in_

("But then what about super-skaterhax?" super-skaterhax isn't all that reliable so like dropping that still makes sense)

mirrormirrorontheweb:
> super-skaterhax isn't all that reliable so like dropping that still makes sense
how is it not reliable?

_pages/en_US/include/3ds-11.17.txt
```diff
@@ -2,5 +2,4 @@
 
 + If your console is running *Luma 10.2.1 or higher*, it is *100% safe* to update. You can check your Luma version by holding (Select) while booting your console.
 + If your console is on an older Luma version, you should [update Luma](checking-for-cfw) before you update your console to 11.17.0.
-+ If your console does not yet have custom firmware, **do NOT update to the latest version** unless this guide instructs you to. The methods on earlier firmwares are easier and more stable.
-+ If your unmodified console is ALREADY on 11.17.0, this guide has been updated accordingly. Both [New3DS](installing-boot9strap-(super-skaterhax)) and [Old3DS](installing-boot9strap-(mset9)) now have free software methods that (for the most part) do not require external hardware or software.
++ If your console does not yet have custom firmware, you can install it now on all firmwares with this guide.
```
lifehackerhansol, line 5: I think it's been long enough that this include could be removed entirely

lifehackerhansol (changes requested):

_pages/en_US/include/3ds-11.17.txt
Gruetzig: dont disagree, but maybe it doesnt belong into this pr

Gruetzig: Would be cool. Just added the change because super-skaterhax is gone with this

SirGamers:
I disagree with removing the alt exploits cuz if MSET9 doesn't work and you can't do Soundhax, what are you supposed to do

Gruetzig:
the methods dont stop working when theyre not on 3ds.hacks.guide. Weve recommended alternate exploits on the support discord before that page existed and theres nothing stopping us from doing so in the future. I think its enough of an edge case to not need it on the main guide. theres the wiki with a lot of exploits that aren't/weren't/probably won't be on the main guide in the future.

SirGamers:
> the methods dont stop working when theyre not on 3ds.hacks.guide. We've recommended alternate exploits on the support discord before that page existed and theres nothing stopping us from doing so in the future. So I think its enough of an edge case to not need it on the main guide. Theres the wiki with a lot of exploits that aren't/weren't/probably won't be on the main guide in the future.

ok then why keep kartdlphax?

_pages/en_US/include/3ds-11.17.txt
lilyuwuu: I think the include should be kept but removed from keyinfo maybe?

_pages/en_US/include/addproxy.txt
```diff
@@ -12,4 +12,4 @@ A visual guide is available for this section [here](/images/screenshots/set-prox
 1. Click OK, then click Save
 1. When prompted, click "Test" to perform the connection test
     + The test should succeed
-1. Click "OK" to continue
+1. Click "OK" to continue
```
lilyuwuu, line 15: this is the coolest change of all time

_pages/en_US/installing-boot9strap-(mset9).txt
```diff
@@ -14,7 +14,7 @@ title: "Installing boot9strap (MSET9)"
 
 ### Compatibility Notes
 
-This method requires a computer running Windows, Linux, or macOS. Phones or tablets are NOT currently supported. If you do not have access to a computer, you will need to use an [alternate exploit](alternate-exploits).
+This method requires a computer running Windows, Linux, or macOS. ChromeOS and Android devices have an app that is currently not covered here. iOS devices are not supported. If you do not have access to a computer, you will need to use an [alternate exploit](https://wiki.hacks.guide/wiki/3DS:Alternate_Exploits).
```
lilyuwuu, line 17: do we want to have this as [an app](link) in the meantime?

lilyuwuu (commented):
as stated in guide discussion i think we should keep the skaterhax page (but not necessarily route it) since it is still usable (keep in mind on NH we mostly see the unsuccessful cases, there are probably a lot of cases where skaterhax works just fine and we just cannot reproduce it for whatever reason)

everything LGTM but i think we should hold off until we have the android/chromeOS page ready, since i don't think mset9 should replace everything until we get to that point

_pages/en_US/include/addproxy.txt
lifehackerhansol: Pretty sure this adds a newline to EOF, I don't particularly mind that because realistically it should be there

_pages/en_US/installing-boot9strap-(mset9).txt
lifehackerhansol: How would you word it?

"An Android application exists for this process; for further support, ask for help at [Nintendo Homebrew on Discord](https://discord.gg/MWxPgEp)." ?

irowiki:
> as stated in guide discussion i think we should keep the skaterhax page (but not necessarily route it) since it is still usable (keep in mind on NH we mostly see the unsuccessful cases, there are probably a lot of cases where skaterhax works just fine and we just cannot reproduce it for whatever reason)
> 
> everything LGTM but i think we should hold off until we have the android/chromeOS page ready, since i don't think mset9 should replace everything until we get to that point

I almost feel like routing n3ds to Skater first and then if it fails, have a quick troubleshooting outline (like .skater) for them to check, and if it still doesn't work, route them to mset9.

This takes care of presumably most people just sailing through skater anyway.

lilyuwuu:
Kinda same, honestly. The main thing stopping me is that routing people to other methods when the current method doesn't work isn't really ideal, so it comes down to basically how reliable we think skater is (like, we don't really know for sure how many people do skater and then have no problems)

irowiki:
> routing people to other methods when the current method doesn't work isn't really ideal

Yeah I was actually editing my post when I saw your reply. Making people possibly waste time with one exploit when they could have done another in roughly the same amount of time is definitely not ideal. 

> so it comes down to basically how reliable we think skater is (like, we don't really know for sure how many people do skater and then have no problems)

Did google analytics get fixed to the point we could track how many people land on skater and then continue onto finalizing? 

I think MSET9 is more reliable than Skater, besides the people needing to do a fresh install (Danny has been talking about changing the script to make a fresh install done by default). Perhaps route it in such a way that MSET9 is default for N3DS but if you have an iDevice or just can't use python, suggest Skater as an alternate exploit

lilyuwuu:
Ok I just looked through guide analytics, and here's what it's looking like.
- Over the past month, about 12,000 users used either MSET9 or super-skaterhax. (6700-5400 split).
- Of the 6700 MSET9 users, ~1600 (31%) make it to Finalizing Setup on their first pass-through, while ~950 (14%) find their way to Troubleshooting. From those that go to Troubleshooting, about 500 (52%) make their way to Finalizing Setup.
- Of the 5400 super-skaterhax users, ~1800 (39%) make it to Finalizing Setup on their first pass-through, while ~700 (13%) find their way to Troubleshooting. From those that go to Troubleshooting, about 300 (42%) make their way to Finalizing Setup.

Comparing these statistics to the methods we consider to be more 'reliable', SSLoth-Browser and Soundhax:
- Over the past month, about 3,200 users used either SSLoth-Browser or Soundhax (1800-1200 split).
- Of the 1800 SSLoth-Browser users, ~650 (36%) make it to Finalizing Setup on their first pass-through, while ~300 (17%) require troubleshooting, of which ~150 (50%) are successful
- Of the 1200 Soundhax users, ~500 (42%) make it to Finalizing Setup on their first pass-through, while ~80 (7%) require troubleshooting, of which ~40 (50%) are successful

In summary:
| **Method**          | **Initial Success** | **Troubleshooting** | **Successful Troubleshooting** | **Total Successful** |
|---------------------|---------------------|---------------------|--------------------------------|----------------------|
| **MSET9**           | 31%                 | 14%                 | 52%                            | 38%                  |
| **super-skaterhax** | 39%                 | 13%                 | 42%                            | 45%                  |
| **SSLoth-Browser**  | 36%                 | 17%                 | 50%                            | 44%                  |
| **Soundhax**        | 42%                 | 7%                  | 50%                            | 45%                  |

- Initial success: Soundhax > super-skaterhax > SSLoth-Browser > MSET9
- Requires troubleshooting: SSLoth-Browser > MSET9 ~= super-skaterhax > Soundhax
- Troubleshooting is successful: MSET9 ~= SSLothBrowser ~= Soundhax > SSLoth-Browser
- Ultimately successful: super-skaterhax = Soundhax ~= SSLoth-Browser > MSET9

Therefore we should definitely keep super-skaterhax. (Success rates for MSET9 likely impeded by lack of command line access / command line being scary, though.)

TheMasterOfMike:
What would the plan be for N3DS 11.14/11.15 then?

They can't do super-skaterhax without updating, so what would be the better approach here:
- just do MSET9 automatically
- have people update, then do super-skaterhax

lifehackerhansol:
> What would the plan be for N3DS 11.14/11.15 then?
> 
> They can't do super-skaterhax without updating, so what would be the better approach here:
> 
> * just do MSET9 automatically
> * have people update, then do super-skaterhax

MSET9

lilyuwuu:
maybe for mset9 we should have a landing page at the beginning and have people select their computer type lol

i feel like the orange message box might not be visible enough

TheMasterOfMike:
> maybe for mset9 we should have a landing page at the beginning and have people select their computer type lol
> 
> i feel like the orange message box might not be visible enough

that would probably be the best approach if there’s a desire to add the Android app and whatever as well

_pages/en_US/installing-boot9strap-(mset9).txt
lifehackerhansol: We should handle this in #2402

lifehackerhansol (approved):

PR #2394, opened 2023-11-10
ssloth-browser: bring back DNS server

lifehackerhansol:
**Description**

The DNS server is restored and functions exactly as the original version of this page intended.

safecerthax remains on the proxy for now as it is not set up on the DNS. Merge includes accordingly.


lifehackerhansol:
This isn't happening anytime soon

Issue #2388, opened 2023-10-29
Use MSET9 instead of Seedminer


Anonymous941:
MSET9 is compatible with every version that Soundhax doesn't run on, and is no longer in beta.  While it isn't as easy as some of the other exploits as it requires exact instructions to be followed and doesn't work on macOS, it's still much better then Seedminer for a number of reasons:

* Time: Seedminer has a huge amount of more steps to be preformed when chained with BannerBomb3 and menuhax.  This causes a long installation time on both the computer and the console
* Processing power: Seedminer requires brute-force power that isn't huge, but also certainly isn't trivial, especially on old computers
* Ease: while MSET9 requires not pressing the touch screen or any buttons to be pressed, Seedminer has a lot of steps that could be difficult to preform
* Requirement of Internet: Seedminer requires that a friend code be added on the 3DS, which requires a working Internet connection on the console and the console being unbanned
* Requirement of upgrading: this isn't a problem of Seedminer, but it's still worth mentioning (and I don't think needs a separate issue): on the New 3DS, you are required to upgrade to the latest firmware if you have most firmware versions, which requires Internet and also has a small chance of causing a brick, especially if CFW was installed before

For these reasons, I think it would be better to recommend MSET9 by default, and then link to the Seedminer methods as an alternative

lilyuwuu:
Honestly not totally against this. Some preliminary thoughts (I'm gathering some thoughts over in #guide-meta on Nintendo Homebrew, and of course anyone else is free to comment here too):

- MSET9 certainly has matured enough and is generally faster and less annoying for us to deal with
- Bruteforcing doesn't have to be done by the end-user through miner services like [BFM](https://seedminer.hacks.guide) and [MovableQ](https://cm360.xyz/movableq/) (#2381)
- MSET9's biggest current drawback is OS support (macOS, which can be worked around, and iOS/iPadOS which most likely cannot)
    - About half of guide visits come from mobile, though we can't know how many people are doing things on a computer and just following along on their phone or whatever
- Seedminer requires Internet, while Seedminer (Mii) doesn't. That being said Seedminer minus the Mii will be dead in \*checks clock* five months anyway.
- I guess there's a hypothetical chance of bricking from a system update, but that's just... pretty much every system update ever, and if it wasn't something that could be restored via Safe Mode there was probably some other issue anyway. Basically I don't think we're going to end up in a Switch 17.0.0 situation sort of thing
[...]

PR #2398, opened 2023-11-16
Add a note about setting correct system date for super-skaterhax

DJ-Laser:
**Description**
Last week I was approached by a few clients (read: friends) who were interested in my modding service. I ran into an issue modding the first one though, super-skaterhax didn't work because the date was wrong. Eventually after trying seven times I looked at the troubleshooting and fixed the date. The second 3DS also had this issue, but I caught it because the first one was still fresh in my mind. I think this info about the date should be right there on the page, to make it easier for people to do this exploit and reduce frustration.


lilyuwuu:
Not conceptually against this change but worth noting that Skaterhax itself is very very picky and so there is an ongoing discussion of forgoing it in favour of [MSET9](https://3ds.hacks.guide/installing-boot9strap-(mset9)). When it works skaterhax is great but there are weird instances where it doesn't work (hence why the troubleshooting is kind ofj ust throwing stuff at the user and hoping it works rather than being a concrete/consistent solution)

DJ-Laser:
This is basically copied from the troubleshooting section, so maybe an include should be used, but idk how to do that.

DJ-Laser:
> Not conceptually against this change but worth noting that Skaterhax itself is very very picky and so there is an ongoing discussion of forgoing it in favour of [MSET9](https://3ds.hacks.guide/installing-boot9strap-(mset9)) [#2388, related PR #2393]. When it works skaterhax is great but there are weird instances where it doesn't work (hence why the troubleshooting is kind ofj ust throwing stuff at the user and hoping it works rather than being a concrete/consistent solution)

 Good point, however an incorrect date is guaranteed to make the exploit fail, so I think we should add a note about it to prevent confusion while super skaterhax is still being reccomended

PR #2416, opened 2023-12-13
Encourage NAND backups more

Anonymous941:
**Description**

This mentions making a NAND backup to avoid bricking in the FAQ


lifehackerhansol:
Thanks for your pull request.

I'm going to say this doesn't help the FAQ in question. Any user reading this FAQ likely doesn't even know what a NAND backup is.

Also, a NAND backup doesn't avoid bricks. It sure makes it difficult to fix, but that doesn't mean it prevents it from happening. It's a minor nitpick, but still, the way it's written now, it isn't a true statement. Note that this question was specifically in place for whether the console will brick during installation of custom firmware, and at that point users won't have a NAND backup anyway.

PR #2427, opened 2024-02-04
finalizing: switch to file-management scriptrunner

lilyuwuu:
# Finalizing Setup: Switch to file-management scriptrunner

## What this PR does

This PR changes the flow of Finalizing Setup. At the moment:

- Users are told to download a file pack, `finalizing.zip`, and to copy the contents of the file pack to the root of the SD card, merging/replacing any existing folders.
- This file pack contains a pre-created `/luma/payloads/` folder with GodMode9 inside, a pre-created `/gm9/scripts/` with the script itself inside, and a `finalize` folder containing all script dependencies (CIAs, donor database, etc).
- Upon holding START on boot, user enters GodMode9, is prompted with essential files backup/RTC, navigates to scripts menu, selects script, lets script run.
- Thus, IF the user successfully copies the contents to root of SD, then hypothetically everything is smooth sailing, barring SD card oddities (hello error 14a).

The main problem is that it is difficult for users to merge SD card files. Some operating systems, like macOS, don't even give you the option to merge two folders together unless you hold a key while dragging-and-dropping. Thus, at best you end up with `luma (1)` or `(gm9 (1)`, and at worst you end up with potential data loss.

With this change:

- Users are told to download two files: `finalize_helper.firm` and `finalize.romfs`
- The file `finalize_helper.firm` will go in `/luma/payloads/`, which has to be created by the user
- The file `finalize.romfs` is intended to go on root of SD, but common user error is accounted for (A `finalize.romfs` erroneously placed in `/luma/payloads/`, `luma`, `Nintendo 3DS`, `3ds`, and `DCIM` will be caught and moved automatically without user propmting).
- Upon holding START while turning on the console, the user will be prompted with the Finalizing Setup Helper Script, which will automatically run.
- The script only does the following things: Check if SD card is writable; ensure `finalize.romfs` is correct; unpack `finalize.romfs`; copy contents of `finalize.romfs` to `SD:/finalize`.
- If these things are successful, the scriptrunner will delete itself, copy GodMode9 to `SD:/luma/payloads/`, and the sequence continues as it does now (user enters GodMdoe9, is prompted with essential files backup, etc)

## Benefits

- Lower chance of user error relating to file extraction (e.g. `SD:/finalizing/finalize`, `/finalize/gm9`, `/finalize/luma`) or folder merging (`luma (1)`)
- **`finalize.romfs` can be checked for file integrity** (currently via hardcoded checksum), which can lower chances of script failure due to file corruption (since it would be more likely to be caught during this file copy step than later on)
- The scriptrunner handling file extraction means that it is _no longer feasibly possible_ for users to be missing files, meaning that errors related to missing files (e.g. missing `SD:/finalize`, missing donor database) can be removed 
- Knowing that all users will have the same files/folder structure (lest the scripts fail in an unexpected way) means that we can reliably add visuals to the script, which can aid with language-agnostic self-serve troubleshooting: 
image

## Disadvantages

- While the chances of user error relating to file extraction are greatly reduced, the chances of user error relating to a misnamed or misplaced `payloads` folder greatly increase
    - Can be resolved if Luma3DS is made to generate an empty correctly-spelled `payloads` folder on boot, since Luma will automatically boot once B9S is installed and Finalizing Setup takes place after B9S is installed
- Transparency: It is less clear to the user what the provided files are doing
    - While these changes greatly assist users who struggle with file management, technical users may be dissuaded by files and folders effectively becoming more 'black-boxed' (while these files are not encrypted or inaccessible in any way, it is not necessarily trivial for someone unfamiliar with the 3DS to extract)
- Maintenance burden is slightly higher, due to the scriptrunner and RomFS requiring a rebuild whenever a change is necessary
    - While technically possible to automate, [I probably won't](https://xkcd.com/1205/)

### Finalizing Setup
- Removal of `finalizing.zip`; addition of `finalize.romfs` and `finalize_helper.firm` to What You Need
- Prompt user to copy `finalize.romfs` to root of SD, and `finalize_helper.firm` to `/luma/payloads/`, with prompt to create folder if not existing
- Update SD file structure images accordingly (`finalizing-root-layout` adjusted to contain just `Nintendo 3DS`, `boot.firm`, `boot.3dsx`, and `finalize.romfs`; new file `finalizing-luma-payloads` depicting `finalize_helper.firm` in `/luma/payloads/`)
- In-line troubleshooting for if the scriptrunner does not launch (accidental SELECT on boot; misspelled `payloads` folder)
- Specify that if the main script (`finalize.gm9`) fails, to restart from Step 3
    - "From this point forward, you can access GodMode9 by holding START while powering on your console" is added to that step to account for this, and also teaches users that holding START on boot triggers GodMode9

### Troubleshooting
- Remove errors # 03, # 09, and # 11 from troubleshooting
    - These errors no longer exist in the script as they all relate to the user having missing or incorrect files, which should no longer occur
- Add errors related to the scriptrunner: # 22 (invalid checksum), # 23 (special case of resolving user error), # 24 (write protection detected)

## Other, non-script related changes made in this PR (to be debated)
- **CFW Check removed from Finalizing Setup** (Section numbers adjusted accordingly)
    - This change was intended to be temporary in the aftermath of 11.17, where O3DS users who updated to 11.17 without CFW would end up without a method. This is no longer the case, so the section no longer serves its use
    - While I recognize that it has found a renewed use as an extra troubleshooting safeguard, I do not feel that this safeguard is necessary and constitutes an unnecessary section for the vast majority of people following the guide
    - At the very least, the line "If custom firmware is not installed by this point for any reason, the next section may render your console unable to be hacked without external hardware." is no longer accurate and should definitely be removed
    - An alternative, less conclusive check could be to add "make sure that System Settings says `Sys.` before the version", since Show NAND or user string in System Settings is enabled by default
    - This is of course open for discussion and admittedly should probably in another PR but here we are
- Move wiki CTA/infobox above 'information and notes'
    - This should hopefully improve visibility of the wiki - while the wiki definitely contains Information and thus might be better suited under the header, that header might also reduce visibility as people don't scroll too far below 'done'
 
## Discussions to be had

- **Are there any unforeseen consequences from this change?** While this sequence appears to work fine for myself and those who have tested it, there is always the potential for logical flaws I might be neglecting
- I feel like the wording "Open the `luma` folder and create a folder named `payloads` inside, if it does not already exist" is a bit messy - any alternative suggestions that still improve clarity upon what we currently do elsewhere on the guide ("copy to `/luma/payloads/`, creating folder if not already exist")?
- Should `finalize.romfs` really be copied to SD root? As-is, the script can detect if `finalize.romfs` is misplaced. Having the user copy `finalize.romfs` to `/luma/payloads/` would make it easier for users to copy files to the right place (since there is just one place instead of two), but miscommunicates that `.romfs` files are payloads when they aren't

By the way I haven't actually tested if these changes actually build correctly, but the site builds so it totally should be fine probably LOL


irowiki:
>CFW Check removed from Finalizing Setup

This will work, if they make it here without CFW they'll find out quick when GM9 doesn't load, perhaps a snippet there instead.


I also support having Luma just create the payloads folder

ihaveamac:
Section IV, Step 2:
> Press and hold (Start), and while holding (Start), power on your console. This will launch the Finalizing Setup Helper

This section should account for people who instead go to the chainloader menu by telling them what to pick, because plenty of people sometimes re-do Finalizing Setup.

PR #2437, opened 2024-03-10
Move some pages to the wiki that aren't directly hacking-related

Gruetzig:
Merge? Maybe wait until translation stuff solved on wiki side
Maintaining two versions of the pages is shit and the wiki is more up-to-date so why not go there?
this does
- Replaces Dumping and GodMode9 Usage guide pages with redirect to its respective wiki pages
- Replaces all links to the aforementioned pages with wiki links
- Deletes SD card checking pages
- Replaces all links to SD card checking pages with wiki page
- Deletes all SD card formatting pages
- Replaces all links to SD card checking pages with wiki page

I figured the Dumping Titles and Game Cartridges and GodMode9 Usage guides were linked a lot in the past so we better not make them 404

Should GodMode9 page stay in the top bar and redirect? Using link directly in the YAML file made link to `https://3ds.hacks.guide/https:/wiki.hacks.guide/wiki/3DS:GodMode9/Usage` so that doesn't work and linking the redirect page makes it so it goes to an empty page for a short time which isn't optimal in my opinion.

### DIFF
--- a/_pages/en_US/ctrtransfer.txt
+++ b/_pages/en_US/ctrtransfer.txt
@@ -31,19 +31,24 @@ To download the CTRTransfer images on this page, you will need a torrent client 
 
 {% include_relative include/ctrtransfer-prep.txt %}
 
-#### Section II - CTRTransfer
+#### Section II - NAND Backup
+
+1. Press and hold (Start), and while holding (Start), power on your console. This will launch GodMode9
+{% include_relative include/nand-backup.txt %}
+
+#### Section III - CTRTransfer
 
 {% include_relative include/ctrtransfer-main.txt %}
 
-#### Section III - Launching FBI
+#### Section IV - Launching FBI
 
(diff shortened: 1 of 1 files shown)


lifehackerhansol:
I think we need a notice that it will be redirected to a different site, rather than redirecting willy-nilly. (Yeah it's still hacks.guide, but still not the same 3ds guide page...)

I can handle that at some point.

Gruetzig:
Only the GodMode9 Usage page when going over top bar or in general and the Dumping Titles and Game Cartridges page as well?

lifehackerhansol:
Any redirect stub we have currently, that redirects outside of 3ds.hacks.guide, should have a notice.

It shouldn't be difficult to do, and we can create an include for it.

Gruetzig:
damn i forgot this existed anyway

PR #2439, opened 2024-03-17
Update f3-(linux).txt

TiRoCode:
**Description**

- ./fwrite changed to ./f3write

While using the guide, I noticed that the command had an error.

PR #2440, opened 2024-03-17
Update f3-(linux).txt

TiRoCode:
**Description**

- replaced h2testw in the first sentence

I rewrote the end of the first sentence because the wrong program was mentioned there.


Gruetzig:
same as #2439

PR #2445, opened 2024-04-02
Changes to Section 2 Step 5 for MSET9

irowiki:
To make things simpler and confuse users less, have users eject on page one without going hunting for mii maker, if mii maker is missing they will need to do Fresh anyway

This avoids people pressing buttons which can lead to a crash regardless


lifehackerhansol (approved):

lilyuwuu (approved):

lilyuwuu:
oops im late

PR #2448, opened 2024-04-08
Add instructions for finalizing error #08

Anonymous941:
**Description**

While helping someone mod their 3DS, they ran into error <span>#</span>08: Dummy Title Database, and I realized it wasn't in troubleshooting.  This PR adds that in English, I don't know other languages so I'd appreciate translations

### DIFF
--- a/_pages/en_US/troubleshooting.txt
+++ b/_pages/en_US/troubleshooting.txt
@@ -448,6 +448,13 @@ The steps below can be attempted in any order, but are listed from easiest to ha
 <details>{{ compat | markdownify }}</details>
 
 {% capture compat %}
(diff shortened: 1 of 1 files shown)


lilyuwuu:
it's not in troubleshooting because the error shouldn't occur anymore, it's usually specifically from following anton retro's guide because they forked the script

not against specifically adding it but i dunno if it would be better to just say "redownload the script cus ur guide is outdated"

lifehackerhansol:
I'm leaning towards no.

The error code is gone from finalize script entirely, no point.

Anonymous941:
Agreed, I wasn't aware that someone I was helping used an outdated script

PR #2464, opened 2024-07-23
super-skaterhax guide refactor

Naim2000:
**Description**

Incorporates many frequent occurrences directly into the guide steps, including the date, region, multiple browser tabs open, etc. (Not really etc.) Also uses @ToxicAven 's super-skaterhax helper. I forgot about that.


_pages/en_US/installing-boot9strap-(super-skaterhax).txt
```diff
@@ -42,12 +42,26 @@ In this section, you will copy the files needed to trigger both super-skaterhax
...
 
+<!-- new section header here? -->
+
+1. Reinsert your SD card into your console
+1. Power on your console
+1. Launch System Settings
+1. Select "Other Settings"
+1. Navigate to "Date & Time" -> "Today's Date"
+1. Tap the Up/Down arrows to set the correct Day, Month and Year
+1. Return to "Other Settings"
```
lilyuwuu, line 57: 'select ok to confirm' is still necessary here so that people don't back out without saving changes
we might also want to say something for the people that already have the date properly set

_pages/en_US/installing-boot9strap-(super-skaterhax).txt
```diff
@@ -56,21 +70,9 @@ In this section, you will visit the browser exploit webpage, which will launch t
 If you haven't already, make sure you have a working Internet connection set up on your console.
 {: .notice--info}
 
-{% capture set_date %}
```
lilyuwuu, line 59: another possibility is to just to have this box as a normal infobox (and not a collapsed-by-default one). Though this would also make the sequence for ensuring correct region settings less smooth.

lifehackerhansol (approved):

PR #2465, opened 2024-07-28
Adding the "Fixing DS mode" section into the "Region Changing" guide

GabubuAvailable:
**Description**
Heyo. I came here in order to help you and every people who plans to fully region change the console to another region.
The reason of doing this PR is because it's missing the DS mode part. You see, When you try to update the network settings on DS mode from the System Settings, or try to load TWiLight Menu++ or a regular DS game, it crashes and gives you this error shown here:

So what I've added in the "Region Changing" guide will cover that part too. For fixing that, [TWLFix-CFW](https://github.com/MechanicalDragon0687/TWLFix-CFW) by MechanicalDragon0687 is required in order to make the DS mode work properly and play DS games on the 3DS/2DS:

If you have any questions or sugestions on how can I make it better, let me know and we'll find a solution!

Thank you,
Gabubu

### DIFF
--- a/_pages/en_US/include/ctrnand-dsmode.txt
+++ b/_pages/en_US/include/ctrnand-dsmode.txt
@@ -1,0 +1,6 @@
+This section will fix the DS mode from crashing once you try to update the network settings or running DS games on your 3DS/2DS.
+
+1. Open the "Homebrew Launcher" then load "TWLFix-CFW"
+1. Press (A) to begin deleting the DS system files from the previous region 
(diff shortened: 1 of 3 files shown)


lifehackerhansol:
I feel as though this is a causation != correlation situation. A region change in and of itself shouldn't break DS mode; I can't reproduce this, for one, and there'd be a lot more complaints if this was a commonplace issue.

We have a troubleshooting section for the DS mode issue itself though, if it really becomes an issue it would be better to just link to that, but for now this seems unnecessary.

PR #2473, opened 2024-08-23
Move torrent downloads to What You Need

TophattedWasTaken:
Waiting on a consensus from #guide-meta (and I probably messed up the ordering for the downloads)


lifehackerhansol (approved):
lily ACKed it

PR #2481, opened 2024-09-23
Update installing-boot9strap-(super-skaterhax).txt

shinobijolteon-alt:
rephrase a step where it says to select on top screen to say select at top of bottom screen to avoid confusion

**Description**


irowiki:
This should probably be added to this PR instead:

https://github.com/hacks-guide/Guide_3DS/pull/2464

PR #2489, opened 2024-09-28
add kor tid & remove extraneous angle bracket

fyredragon69:
**Description**

this pull request brings world peace by adding the kor tid for home menu theme data and removing an extraneous angle bracket


lifehackerhansol:
(by the time I saw this I already fixed that angle bracket thing, so that is omitted from the PR)

PR #2498, opened 2024-10-04
Automatically replace link to file accordingly to selected version

danny8376:
Automatically provide correct soundhax/super-skaterhax files when navigated from version selector.

NOTE: this breaks translations


lifehackerhansol:
After we temp hack the translations there I think this is alright

lifehackerhansol (approved):

PR #2500, opened 2024-10-08
troubleshooting: split into separate pages

lifehackerhansol:
**Description**

Under this new structure, troubleshooting pages are now split into respective guide pages. With SafeB9SInstaller issues as an include step, this allows to have a complete troubleshooting section for each guide page, shrinking the previously-massive troubleshooting page into understandable fragments.

TODO:
- [x] add all guide pages that have inline troubleshooting
- [x] fix inline troubleshooting to link to new pages


lifehackerhansol:
Test website for reference: https://3dsguide.lifehacker101.net/

PR #2504, opened 2024-10-13
Added CFW Check before Get Started

AdRoz78:
Adding CFW Check before Getting Started, as many people skipped the CFW Check paragraph.


Gruetzig (changes requested):
This PR happens to completely remove key information from the guide path.
Solution: Revert the changes in index and link from Key Information to Checking for CFW instead.

lifehackerhansol:
If we are doing this, it should be added to the sidebar.

Check `docs/.viteprese/i18n/en_US.js`

docs/get-started.md
```diff
@@ -6,29 +6,12 @@ head: [
 ]
 ---
 
-# Get Started
```
lifehackerhansol, line 9: This header should not be removed, it is the title of the page.

lifehackerhansol (changes requested):

PR #2538, opened 2025-01-31
consoleVersionSelect: add version prefix

ManiacOfGitHub:
Screenshot 2025-01-31 143517
If they don't select "Ver.", they have Luma3DS and have skipped the CFW check.

Hopefully, I didn't screw up any translations with my "otherPrefix" thing. Also, I wasn't 100% sure what to call the Ver/Sys/Emu/etc. thing, I think version prefix makes sense though.

### DIFF
--- a/docs/_internal/consoleVersionSelect.html
+++ b/docs/_internal/consoleVersionSelect.html
@@ -15,7 +15,7 @@
       <option>Ver.</option>
       <option>Sys.</option>
       <option>Emu.</option>
-      <option>{{ $frontmatter.otherPrefix }}</option>
+      <option>(Other)</option>
     </select>&nbsp;<select id="major">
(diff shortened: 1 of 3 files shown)


lifehackerhansol:
I like this, but I'm pretty certain this kills translations.
Can you temporarily hardcode "(Other)" in the HTML directly, we can use the frontmatter string when Crowdin has updated everything.

ManiacOfGitHub:
Looking back on this, I feel like we should mention that they already have CFW. Maybe a new page that explains that you cannot follow the main path while modded, and also has them do the CFW check again and if they still go to the HOME menu, then send them to the Discord server because a button might be broken or something.

lifehackerhansol (approved):

PR #2545, opened 2025-03-01
Update installing-boot9strap-(mset9-cli).md | SECTION I POINT 2 MADE CLEARER

onionxenzia:
the point sounded like if you should delete all existing files and directories, i think it is clearer now specifying that your existing directories should remain


lilyuwuu:
Thanks for the PR.

The term "overwriting any existing files" is used throughout the guide and I think it is generally clear enough. Adding more wording might just make the text more confusing / unwieldy.

Leaving this open in case anyone has any other comments

lifehackerhansol:
This PR is already adding more confusion as far as I can tell.
If we _had_ to change something, I'd just add "if asked" to the end of that sentence.

If we're going with that, make sure to add it to every occurrence of this sentence (which is basically everywhere)

lilyuwuu:
Closing due to lack of activity

PR #2547, opened 2025-03-16
Update super-skaterhax to include a warning after step 14 (due to the recent SSL cert change)

allxvarl:
Updates briefly the first warning at the top of the page adding:

"If you have been redirected here form a video guide or a link, do not continue.

If you ignore this warning, you will get an error when pressing the "GO GO!" button."

and after step 14, duplicates the warning to add 

"If you get the error "An error has occurred. Please save your data in any software currently in use, then restart the system." after pressing GO GO! :

This method is currently not working due to server-side SSL certificate changes. The timeline for a fix is unknown. In the meantime, follow Installing boot9strap (MSET9) instead."

### DIFF
--- a/docs/installing-boot9strap-(super-skaterhax).md
+++ b/docs/installing-boot9strap-(super-skaterhax).md
@@ -14,7 +14,7 @@ This method is currently **not working** due to server-side SSL certificate chan
 
 If you have been redirected here form a video guide or a link, do not continue.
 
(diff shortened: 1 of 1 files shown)


allxvarl:
Fixed my bad english, i wrote "when" instead of "after"

allxvarl:
Closing due to https://github.com/hacks-guide/Guide_3DS/commit/a615e79af8ec9ad399e69c431e2f196d45722de0 since it's pretty much the same thing

PR #2560, opened 2025-04-19
mset9-*: sync some text

lifehackerhansol:
**Description**

Particularly on-console steps don't really have any reason to be different between mobile and desktop versions.

### DIFF
--- a/docs/installing-boot9strap-(mset9-play-store).md
+++ b/docs/installing-boot9strap-(mset9-play-store).md
@@ -2,7 +2,7 @@
 
 ::: details Technical Details (optional)
 
-[MSET9](https://github.com/zoogie/MSET9) is an exploit for the System Settings application developed by [zoogie](https://github.com/zoogie). It exploits a flaw where the ID1 (the second 32-character folder name inside of the Nintendo 3DS folder, inside of the ID0) can be *any* name as long as it is 32 characters. Performing a specific sequence of actions results in the console executing the instructions that are encoded into the ID1 folder name, which can be used to grant full control over the 3DS.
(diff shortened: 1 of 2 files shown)


docs/installing-boot9strap-(mset9-play-store).md
```diff
@@ -22,17 +22,17 @@ On Android phones/tablets, the minimum Android version required is 6.0 (Marshmal
...
     * [ZArchiver](https://play.google.com/store/apps/details?id=ru.zdevs.zarchiver)
     * If you wish, you can sideload these applications instead
-* The latest release of [MSET9](https://github.com/hacks-guide/MSET9/releases/latest) (the MSET9 `.zip` file)
 
 ## Instructions
 
 ### Section I - Prep Work
 
-In this section, you will prepare the SD card data necessary for the MSET9 exploit to trigger.
+In this section, you will prepare the MSET9 exploit by **temporarily** creating a new HOME Menu profile with no user data, and then setting up that profile with only the minimum data required for MSET9 to trigger. Your existing user data will disappear, but will come back when you are finished with this page.
```
danny8376, line 35: maybe
``In this section, you will prepare the MSET9 exploit by **temporarily** creating a new HOME Menu profile with almost no user data, and then setting up that profile with only the minimum data required for MSET9 to trigger. Your existing user data will disappear, but will come back when you are finished with this page.``

docs/installing-boot9strap-(mset9-play-store).md
```diff
@@ -58,13 +58,13 @@ In this section, you will prepare the SD card data necessary for the MSET9 explo
 1. Tap on the photo corresponding to your console model, then at the bottom, pick your current firmware version
 1. If you see the prompt of `Hax ID1 Created`, tap OK to continue
     + If you get an error, check the [troubleshooting](troubleshooting-mset9), then try again
-    + Your 3DS will appear to lose most data / no user-installed apps on HOME Menu. This is expected. Your data will come back at a later step
+    + Your console will appear to have no data / no user-installed apps on HOME Menu. **This is expected.** Your data will come back at a later step
```
danny8376, line 61: personally still prefer lose most data, as have no data at all is not correct, nand data is never effected, and this is actually asked a lot

PR #2567, opened 2025-06-02
browserhax: fallback to MSET9 if DS mode is not working

lifehackerhansol:
**Description**

Some devices run into a specific error on exploits that make use of kernelhaxcode_3ds:
- PrepareArm9ForTwl returned error c8804631!

This was traced down to the agbhax exploit chain, which is used under the following conditions:
- NATIVE_FIRM
- 11.4 or newer

However, if the "DS Internet" title is broken or missing, the exploit will fail.

In these cases, fall back to MSET9, which doesn't have any of the above issues, at the expense of convenience.


lilyuwuu (approved):
Maybe change "you should" to just "follow", to be more direct? Otherwise lgtm

Issue #2568, opened 2025-06-09
"Get Started" stuck in a loop for non-modded consoles


Hrshl7:
**Pages with issue(s)**
https://3ds.hacks.guide/get-started.html
https://3ds.hacks.guide/checking-for-cfw

**Description of the issue(s)**
"Get Started" page demands to fill in information about the 3DS in order to know which method to use.
Confirming leads to "Checking for CFW" pages saying "If your console boots to the normal HOME Menu, return to Get Started".
This starts a loop between the two pages.

Also, in the left sidebar "Multiple Options" is not clickable.

danny8376:
if you console have prefix that's not `Ver.` , it's definitely modded, so loop for non-modded consoles won't happen
you might get looping for modded one if you have broken select button or it's stealth luma or whatever rare setup, that probably should be fixed properly though

PR #2569, opened 2025-06-10
Already have CFW

ManiacOfGitHub:
This page is similar to Checking for CFW except it tells you outright that you have custom firmware and redirects you to the Discord server for assistance if you boot to the HOME menu (in cases of broken buttons or Stealth Luma or something).

Why?

1. "Solves" #2568
2. Allows us to get a .alreadycfw command (or something like that) so people are less confused when giving support. (No more "but I already have CFW installed" when sending .cfwcheck)

### DIFF
--- a/docs/already-have-cfw.md
+++ b/docs/already-have-cfw.md
@@ -1,0 +1,52 @@
+# Already have CFW
+
+## Required Reading
+
+Your console already has custom firmware installed. This may not be obvious, as custom firmware is often invisible in day-to-day usage.
+
+This section will help you determine how to upgrade your custom firmware setup.
+
+::: info
+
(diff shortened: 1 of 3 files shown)


lifehackerhansol:
IMO we should instead add a disclaimer at the top of existing checking-for-cfw, "if you landed in this page your console appears to have some funky things installed and we should check, if you're sure nothing happens join nh and complain"

ManiacOfGitHub:
At that point, it would be easier to make it a message that appears on Get Started, as we do [here](https://github.com/hacks-guide/Guide_3DS/blob/6ca1f4d72aa238bc5e369ff2c2087699417321f3/docs/_internal/consoleVersionSelect.html#L123) right? Like make it say: "Your console may already have custom firmware installed. Please retry Section I, and if you still reach the HOME Menu, join Nintendo Homebrew on Discord for assistance."

Not sure how we would avoid breaking translations tho

PR #2570, opened 2025-06-11
Replace guiformat with Rufus in +64GB SD formatting documentation

kengallego:
**Description**

guiformat does not provide an option to select the MBR partition table, which caused issues when formatting SD cards configured with GPT. This commit updates the documentation to use Rufus, which allows selecting the partition scheme (MBR/GPT)

### DIFF
--- a/docs/formatting-sd-(windows).md
+++ b/docs/formatting-sd-(windows).md
@@ -11,7 +11,7 @@ This page is for Windows users only. If you are not on Windows, check out the [F
 ## What You Need
 
 * **For all SD card sizes**: The latest version of [SD Formatter](https://www.sdcard.org/downloads/formatter/sd-memory-card-formatter-for-windows-download/)
-* **For SD cards 64GB or larger only:** The latest version of [guiformat](http://ridgecrop.co.uk/index.htm?guiformat.htm)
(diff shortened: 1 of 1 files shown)


Naim2000:
This is what SD Card Formatter's job is

kengallego:
> I’m not sure if you have read the formatting SD guide for Windows, but we do use SD Card Formatter because it does touch the partition table unlike GUIFormat, converting to MBR and ensuring the proper specs of the SD. Additionally, replacing GUIFormat with Rufus isn’t the best idea despite the fact it can work because there are just simply many configurations on it that can be confusing to most. With GUIFormat, all you have to do is just select the proper cluster size and that’s it.

I followed "Section II" but GUIFormat formatted my SD maintaining my GPT partition untouched. Thus, my 3DS didn't recognize the SD card.

ajl721:
Right, and again, if the SD was already GPT, running GUIFormat solely wouldn’t change that. This is why we run SD Card Formatter first and then follow up with GUIFormat.

VannyBuns:
Guiformat's main purpose is formatting SD cards that are larger than 32GB to FAT32. It was never intended to change partition tables. Like the others said, SD card formatter are already does this.

kengallego:
So then this is no "Section I" or "Section II" but "Section I" and then "Section II"...

PR #2554, opened 2025-04-04
MSET9 CLI iOS

danny8376:
well, MSET9 CLI for iOS

put as draft since MSET9 side is not fully tested (especially for mac), but can confirm it actually work on ios and this can actually be independent from normal mset9 cli (ideally not though)

https://6313c665.dannyaam-3ds-guide-staging.pages.dev/installing-boot9strap-(mset9-cli-ios)
deployed main guide page for this pr if anyone needed

PR #2571, opened 2025-06-30
Add platform warning to base MSET9 page.

TheMasterOfMike:
I noticed while checking the base MSET9 page that platforms like iOS aren't listed, _but_ that the warning about "not having access to any of these devices" only appears _after_ you select the platform on the base MSET9 page, which doesn't necessarily seem ideal (since what platform would you select if you're using iOS on the base MSET9 page? (which doesn't currently have this warning)).

This PR attempts to fix this by simply adding a variant of the same wording to the base MSET9 page.

(Note: It might be better to just say "If your platform isn't listed here" at the start, but as that could _potentially_ be more confusing depending on how somebody looks at that, I'll defer to the other people looking at this PR as to whether or not that wording would make more sense.)

### DIFF
--- a/docs/installing-boot9strap-(mset9).md
+++ b/docs/installing-boot9strap-(mset9).md
@@ -6,6 +6,12 @@ The instructions for MSET9 differ depending on the computer, phone, or tablet **
 
 Choose the operating system of the device **that you will be using to mod your console**. The device will need a way to read and write files onto the 3DS SD card.
 
+::: warning
+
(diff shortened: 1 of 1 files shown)


TheMasterOfMike:
> In my opinion, it should be a footnote rather than a header because after the iOS PR gets merged, there will be basically no modern desktop/mobile OS that isn't supported by MSET9.

Late reply, but imo this PR/change is largely irrelevant if/when the iOS PR gets merged (what relevant platforms don’t fall under the platforms listed at that point?).

That being said, until that PR gets merged, it makes more sense as a header in large part to make sure that the average reader doesn’t miss that section (and also to better align with the other current MSET9 pages)

lilyuwuu:
Redundant now that #2554 is merged. Thanks for the motivation to finally do so :p

PR #2574, opened 2025-07-17
Add workflow for automatically release gm9 scripts

danny8376:
Automatically put docs/public/gm9_scripts into release so can be linked for download that won't be opened as text on browser

Requires read/write permission for actions but guess that's already the case


lifehackerhansol:
At this point I'd rather put it in a separate repository

PR #2579, opened 2025-07-24
skaterhax: small clarity improvements

TophattedWasTaken:


danny8376:
It doesn’t need to be console timezone and i think calculating something like japan timezone only complicate things for some people… (google exists i guess)

danny8376:
> If timezone doesn't matter (guessing it works if you're within 24 hours?) then I think the comment about timezone should be removed all together and should just say current date and time

What you mean for comment about timezone?

If you mean country settings, that’s unrelated to  timezone (it might affect time/date check for ssl cert? but as long as it can be treated valid, timezone doesn’t really matter)

PR #2584, opened 2025-08-13
finalizing-setup.md: add notes about pretendo to finalizing-setup

fyredragon69:
adds brief description about what pretendo is and does, as well as the official guide to the end of finalizing


lifehackerhansol:
forcepushed to fix newlines, in markdown you need to add 2x newlines for it to render as multiple lines which seems to be the intention

PR #2586, opened 2025-08-22
Updated the price of the Ace 3DSX flashcart.

ShinyTillDawn:
The listing on nds-card.com shows a "New price" of $24.99 (US).

**Description**

The pull request updates the price of purchasing an Ace 3DSX ntrboot flashcart from $17.99 to $24.99. It ensures that a user isn't surprised when they see that a flashcart is more expensive than what was listed on the guide.


lifehackerhansol (approved):
unfortunate

PR #2594, opened 2025-09-09
Update ja_JP.json

ss2552:
**Description**


lifehackerhansol:
Thank you for your submission. We don't take translation strings via pull request, but rather from Crowdin, as the Contributing page suggests: https://crowdin.com/project/3ds-guide

I do see that the Japanese translation on Crowdin is nearly 100%. You may want to join the translation Discord server listed in the page above to request to be a proofreader as it likely hasn't been touched in ages.

PR #2600, opened 2025-09-21
treewide: standardize installing boot9strap section

TophattedWasTaken:


TophattedWasTaken:
have we confirmed the issue occurs with updating b9s and atob? this was definitely intentional and I for sure did not forget they existed

lifehackerhansol:
I mean I doubt it actually happens there but I figured "we're standardizing it why not cover our ass and just standardize all the way"

PR #2631, opened 2026-02-28
Making markdown styling more consistent, update README.md and LICENSE.md

grunt22fox:
This PR makes the markdown styling more consistent in line with similar changes made to the Wii guide through https://github.com/hacks-guide/Guide_Wii/pull/183.

More specifically, this PR does the following:

* Updates the copyright date in the license, as well as the guide pages
* Removing unneeded HTML tags and replacing them with Markdown equivalents
* Adding flags to the language listing
* Adding badges to the README and rewording things, making it more in line with the Wii guide's README (this is a subjective change)
* A handful of minor textual changes
* Adding periods to the end of most steps on the guide
* Replacing button indicators that were surrounded in quotations or parentheses with code backticks instead
      * EXAMPLE: Instead of (Left Shoulder) + (D-Pad Down) + (Select), it is now `Left Shoulder + D-Pad DOWN + SELECT`
* Various other markdown changes that I may have forgotten

I omitted the markdown styling guidelines that I drafted on the Wii guide's README, as this is not a guide that I actively contribute to. I would appreciate feedback in the comments about if this is something that might be desired. I also copied over the `markdownlint.json` that I used for linting to accomplish some of these changes but it can also be removed if desired.


Gruetzig:
Hi, I feel like the period thing has barely any benefit and will reset all translations for strings that, which will lead to some chaos. I also think there should be some discussion on the formatting of arrows and the button combination change you proposed. 
I do not know how the arrows look in the final version, so if they are interpreted by the markdown parser and made into proper arrows or just stay >, in which case I'd say the original straight up looks better.
For the button thing, I think wrapping in code does increase readability, but the parantheses around each button should remain. This helps with longer words like Left Shoulder, and I also think it just looks kinda cute to have this round shape around the buttons, esp with the 3DS having circular ABYX buttons.

Other than that, big fan of the change from "" to \`\` and the flags are also cute. c:

PR #2632, opened 2026-02-28
Update NPM packages and Vitepress to latest versions

grunt22fox:
This PR updates NPM packages to their latest versions while making changes necessary to remove errors or warnings that impede guide functionality.

Vitepress 2.0 alpha is tested as working in production on other guides and has made no observable impact on the ability to use or access the site.

As a consequence of updating to the latest packages, this PR will additionally resolve various vulnerabilities that have been accumulated over the past several months (which may not apply to a production deployment but are nice to resolve as part of general maintenance).

The following is a more granular view of the changes made in this PR:

1. Fix vulnerabilities as identified by NPM through package updates
2. Update packages as stated below:
    * Vitepress - 1.6.4 > 2.0.0-alpha.19
    * fontawesome-svg-core - 7.2.0 > 7.3.1
    * free-solid-svg-icons - 7.2.0 > 7.3.1
    * vue-fontawesome - 3.2.0 > 3.3.3
3. Update internal JS files in the `.vitepress` directory to comply with future changes to Vite (transitioning to Node.js for configuration loading) that require explicit declaration of ECMAScript modules. .JS files containing ECMAScript module code were renamed to .MJS (and imports reconfigured) to resolve runtime warnings.

Website is tested working on Fedora 44 with the above applied updates as of 08/25/2026.

As a consequence, this PR will resolve #2659, #2660, #2663, #2665, and #2666.


lifehackerhansol:
I updated the other deps separately, but I'm not a fan of shipping alpha in production precisely due to the name of the version. Last time I did that in production I switched an entire website to a different framework.

Furthermore, neither Vite vulnerabilities identified affect us because we do not use SSR in production. It does affect when running the dev instance, which nobody should be exposing to the public anyway and most certainly shouldn't be used in prod (and we can't even use it in prod because we're using GitHub Pages.)

grunt22fox:
Updated the main PR message accordingly to reflect newer changes made

PR #2649, opened 2026-06-13
Add references to eject info

ccastle14:
**Description**

Adds directions to eject sd card from computer/phone/tablet, and adds clickable links to the key information section on safely ejecting.


Random-Penguin:
Why do some of the entries not link to the relevant section in "Getting Started" like the others?

ccastle14:
> Why do some of the entries not link to the relevant section in "Getting Started" like the others?

If the same numbered list had more than one "Eject SD card" step, I only threw in the help link on the first one, since it seemed unnecessary

Random-Penguin:
Ah, I see it now. Sorry...

lilyuwuu:
I'm not opposed to having a separate step that explicitly tells people to eject their SD card but I don't think the "See [Key Information](key-information#eject-sd-card) if you are unsure how to safely eject an SD card" needs to be repeated so often, at most once per page at the top of the page

PR #2662, opened 2026-07-21
Update installing-boot9strap-(super-skaterhax).md

developer51709:
**Description**
Added a note about changing the language to Spanish when installing boot9strap via the super skaterhax method on USA region consoles.


lilyuwuu:
I can't say I've heard of specifically Spanish increasing the success rate on USA region consoles. The standard advice has always been to cycle between the available languages until it works (or never works, hence the suggestion to switch to MSET9 after ten attempts).

Eviolites:
Personally when super-skaterhax first became available, we were recommending a lot of people switch to Spanish and that seemed to increase success rate, I think it does need to be more thoroughly tested with new helpees to see if this was coincidence or cause.

Edit: changed SSH to super-skaterhax to avoid confusion

